PR #1, opened 2021-10-07
Feature/fewsnet download

Tinkaa:
First function as part of the new food insecurity pipeline. Please be critical as I will continue using this style + I like to learn. 

Few small questions: 
- how do we want to name folders? Do you think `food_insecurity` fits here or should we call it `ipc` (since FN uses ipc but can be confusing compared to ipc global)
- how long should files be? E.g. should `fewsnet.py` be rebranded to `fewsnet_download.py` and have another `fewsnet_processing.py`? Or should that all be in one file?
- do we want a readme for each folder?
- Do we want to make each folder a package, i.e. add an `__init__.py`?


aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+            valid_file = True
+        except Exception:
+            # indicates that the url returned something that wasn't a
+            # zip, happens often and indicates data for the given
+            # country - date is not available
+            logger.debug(f"No zip data returned from url {url}")
+        # remove the zip file
+        zip_path.unlink()
+
+    except Exception:
```
turnerm, line 57: Same as above, could a catch-all exception be avoided?

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+        path of dir to which the zip content should be written
+    use_cache: bool
+        if True, don't download if output_dir already exists
+
+    Returns
+    -------
+    country_data : bool
+        if True, country data for the given date and iso2 exists
+    """
+    base_url_country = "https://fdw.fews.net/api/ipcpackage/"
```
turnerm, line 96: This should be made a constant

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+    url_country = (
+        f"{base_url_country}?country_code={iso2}"
+        f"&collection_date={date.strftime('%Y-%m')}-01"
+    )
+    output_dir_country = output_dir / f"{iso2}{date.strftime('%Y%m')}"
+    # question: is there a better way to define this zip path?
```
turnerm, line 102: I'm not sure, do I understand correctly that you're creating a directory with a `.zip` in the directory name?

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    # question: is there a method to force the input to be upper-cased?
+    # upper case iso2 and region_code since they are upper-cased
+    # in fewsnet's url
+    iso2 = iso2.upper()
+    region_code = region_code.upper()
+
+    # we prefer the country data as this more nicely structured
+    # thus first check if that is available
+    # question: is it a good or bad idea to split this to two functions?
```
turnerm, line 210: Definitely good to split this up! Also, I would make the functions private (add an underscore to the front of their names) if they are only meant to be used within this module

turnerm (commented):
Looks really great! The one main question I have is, which of these functions is meant for the user? It would be useful to provide examples on how they are meant to be used.

aatools/food_insecurity/fewsnet.py
Tinkaa: Not entirely. There is the `output_dir_country` which is a directory that the content of the zip file will be written to. And then there is the zip file itself `zip_path_country`, which has the same name as the directory but has the `.zip` extension because it is a zip file. 
But I think I am trying to optimize it too much while loosing clarity. Will change it to use `stem` instead

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def download_fewsnet(
+    date: datetime,
+    iso2: str,
+    region_name: str,
+    region_code: str,
+    # question: is it good practice to have
+    # output_dir as input arg or define that with a config?
+    # question: is it too much to require it to be a Path object?
```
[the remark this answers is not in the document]
Tinkaa: See above. Added the option to have it as str and convert it to Path if the type is a str

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,212 @@
...
+            # country - date is not available
+            logger.debug(
+                f"No zip data returned from url {url} "
+                f"check that the area and date exist."
+            )
+        # remove the zip file
+        zip_path.unlink()
+
+    except HTTPError:
+        logger.info(f"Couldn't download url {url}")
```
turnerm, line 64: How often does this happen? Would it be useful to implement an auto-retry?

The exception seems a bit general, do you think a more detailed error message would be helpful? i.e. if you do `except HTTPError as e` you can print out `e` via the logger

aatools/food_insecurity/fewsnet.py
```diff
@@ -98,9 +98,9 @@ def download_fewsnet_country(
         f"{base_url_country}?country_code={iso2}"
         f"&collection_date={date.strftime('%Y-%m')}-01"
     )
-    output_dir_country = output_dir / f"{iso2}{date.strftime('%Y%m')}"
-    # question: is there a better way to define this zip path?
-    zip_path_country = Path(f"{output_dir_country}.zip")
+
+    zip_path_country = output_dir / f"{iso2}{date.strftime('%Y%m')}.zip"
```
turnerm, line 102: Do we need to save the zip file, or just the output? What do you think of adding to `download_zip` the option of removing the zip file after extraction? Then you could just save the zipfile to `/tmp`

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,212 @@
...
+    country_data : bool
+        if True, country data for the given date and iso2 exists
+    """
+    base_url_country = "https://fdw.fews.net/api/ipcpackage/"
+    url_country = (
+        f"{base_url_country}?country_code={iso2}"
+        f"&collection_date={date.strftime('%Y-%m')}-01"
+    )
+
+    zip_path_country = output_dir / f"{iso2}{date.strftime('%Y%m')}.zip"
```
turnerm, line 102: Do we need to save the zip file, or just the output? What do you think of adding to download_zip the option of removing the zip file after extraction? Then you could just save the zipfile to `/tmp`

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,212 @@
...
+    output_dir_country = output_dir / zip_path_country.stem
```
turnerm, line 103: What do you think of the following instead:
```python
output_dir_country = output_dir / f"{iso2}{date.strftime('%Y%m')}"
zip_path_country = Path(f"{output_dir_country}.zip")
```
That way you don't need to repeat `output_dir` which I found confusing, and it makes it clear that `zip_path_country` is the same as `output_dir_country` just with a `.zip` at the end

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+
+    try:
+        download_url(url, zip_path)
+        logger.info(f"Downloaded {url} to {zip_path}")
+
+        try:
+            unzip(zip_path, output_dir)
+            logger.debug(f"Unzipped {zip_path}")
+            valid_file = True
+        except Exception:
```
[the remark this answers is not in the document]
turnerm: I think the changes that you made are good!

aatools/food_insecurity/fewsnet.py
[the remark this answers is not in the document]
turnerm: Did you use `isinstance` to check for the type? Then it [should work](https://github.com/python/mypy/issues/8383)

aatools/food_insecurity/fewsnet.py
```diff
@@ -0,0 +1,218 @@
...
+        e.g. "east-africa"
+    region_code : str
+        code that refers to the `region_name,
+        e.g. "EA"
+    output_dir : Path
+        path to dir to which the data should be written
+    use_cache : bool
+        if True, don't download if output_dir already exists
+    """
+    # question: is there a method to force the input to be upper-cased?
```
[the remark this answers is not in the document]
turnerm: Sooo I think what we ultimately want is for each method to create a config object, and that will contain a standardized iso3. We should allow the user to pass it in any case but definitely take care to correct the case from our end. Does that make sense?

aatools/food_insecurity/fewsnet.py
Tinkaa: Got it, thanks!

aatools/food_insecurity/fewsnet.py
Tinkaa: Yeah that is what there was before, but then it coused confusion. But guess you right it is still better than what there is now. Will change it back

aatools/food_insecurity/fewsnet.py
turnerm: You can use the Python [`tempfile`](https://docs.python.org/3/library/tempfile.html) module. Alternatively, I think you can just keep the zipfile in memory while you extract it, see [here](https://techoverflow.net/2018/01/16/downloading-reading-a-zip-file-in-memory-using-python/). Probably not a good idea for large zipfiles though.

aatools/food_insecurity/fewsnet.py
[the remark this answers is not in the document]
Tinkaa: Ohh this is cool, this works, thanks

aatools/food_insecurity/fewsnet.py
Tinkaa: @turnerm changed in [the last commit](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/1/commits/06434e81f46ad3bd4c928c4b6d311773ce1bada9). Can you check it? Then I think I solved all the issues of this PR so if you could check one last time and then approve or not?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -0,0 +1,233 @@
...
+                # indicates that the url returned something that wasn't a
+                # zip, happens often and indicates data for the given
+                # country - date is not available
+                logger.debug(
+                    f"No zip data returned from url {url} "
+                    f"check that the area and date exist."
+                )
+
+    except HTTPError as e:
+        logger.info(e)
```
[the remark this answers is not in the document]
Tinkaa: Wow you are so good, thanks for all this help. I am having a bit hard time imaging all the cases that this might occur. But what I have encountered so far is when the url is incorrect and thus gives a 404. In that case it should indeed exit and then it makes sense to not catch the error so changed that

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] - 2021-10-18
+
+- functionality to download and save FewsNet data
```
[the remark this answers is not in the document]
turnerm: Oh I meant like in the example [here](https://keepachangelog.com/en/1.0.0/), they suggested grouping under "Types of Changes"

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -0,0 +1,227 @@
...
+    url_region_date = BASE_URL_REGION.format(
+        region_code=region_code,
+        region_name=region_name,
+        YYYY=date.year,
+        MM=date.month,
+    )
+
+    output_dir_region = output_dir / f"{region_code}{date.strftime('%Y%m')}"
+    zip_filename_region = f"{output_dir_region.name}.zip"
+    if not output_dir_region.exists() or use_cache is False:
```
turnerm, line 160: Shouldn't this be `and`? (But if checking for the cache, shouldn't you check for the files themselves? Maybe it's tough with this data as you don't know the filenames)

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -0,0 +1,233 @@
...
+
+            try:
+                unzip(zip_file_path=zip_path, save_dir=output_dir)
+                logger.debug(f"Unzipped to {output_dir}")
+                valid_file = True
+            except zipfile.BadZipFile:
+                # indicates that the url returned something that wasn't a
+                # zip, happens often and indicates data for the given
+                # country - date is not available
+                logger.debug(
```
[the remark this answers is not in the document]
turnerm: Ah sorry, I understand the structure better now. You're right that you can leave it as debug, but then I would do the following:
- make sure you set `valid_file = False` in the `except`
- On the very last line of this module, instead of just logging an error, I would raise an error:
```python     
if not region_data:
    raise RuntimeError(f"No data found for {date.strftime('%Y-%m')}")
```

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -0,0 +1,233 @@
+"""
```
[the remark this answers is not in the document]
turnerm: Well, with every new folder you are actually creating a new package. It could be a bit of overkill to have only one module per package. However, if this module grows and splits into multiple files then it makes more sense. 

I think another disadvantage of overly hierarchical structure is then the imports get overly complex. But I [realized](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces) that we are actually suppose to define all public interfaces in `__all__`, so this could be a work around (I'm still trying to get my head around it a bit though)

mcarans:
@Tinkaa Javier messaged me today saying "also just heads up FEWSNET is also coming our way, so i may need your support to review their API, this is a big one, i have been waiting for ages for FEWSNET and fianally it looks they will come...so i will engage you once we reach the stage of API integration"

It looks like you are already using an API from FEWSNET. I'm not sure if Javier is referring to a new API or something else. Have you been in contact with Javier about FEWSNET?

CHANGELOG.md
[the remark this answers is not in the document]
turnerm: Ah sorry, somehow in the inline comments the change was not shown. Thanks Tinka!

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -0,0 +1,227 @@
...
+        e.g. "EA"
+    output_dir : Path or str
+        path to dir to which the data should be written
+    use_cache : bool
+        if True, don't download if output_dir already exists
+    """
+    # upper case iso2 and region_code since they are upper-cased
+    # in fewsnet's url
+    iso2 = iso2.upper()
+    region_code = region_code.upper()
```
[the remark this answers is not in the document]
turnerm: So I guess my reasoning was the following (but no worries if you want to keep it this way, it's a small detail): I thought it's better to pass around the unformatted string, otherwise in every function where it gets passed, you need to specify that it must be a lowercase iso3, because the formatting will not be taken care of later. Just how you pass the datetime object around and only format it at the end, rather than passing the month and year. Hope that makes sense.

Tinkaa:
@turnerm changed the order of functions as discussed. Let me know if you think there is anything else!

Tinkaa:
@turnerm yaay gonna merge it

PR #2, opened 2021-10-07
General setup and example

turnerm:
This PR is intended to to explore possibilities for the structure and tooling of the new repository. It includes:

- many new pre-commit hooks
- Sphinx documentation (linted to numpy style)
- a migration example of the [area](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/develop/src/utils_general/area.py) module from the main AA repo

Other points to note:
- I thought we could try out [pbr](https://pypi.org/project/pbr/) for the versioning. It will automatically use the last tag + a custom addition to generate a version number each time you run setup.py. It also auto-generates the author list and changelog. Perhaps the former should be deleted (or we can create one with our UN emails). As for the changelog, it would require us to keep a clean commit history since it's based on commits, so we may prefer to just do that by hand.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,0 @@
-CHANGES
-=======
-
-0.0.1
------
-- Setup tooling
-- Implemented CDS Area module
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,3 @@ line-length = 79
 [tool.isort]
 profile = "black"
 line_length = 79
-
-[tool.flake8]
-extend-ignore = "SFS301"


mcarans (commented):
This looks great, particularly the use of different requirements files, type hints and Sphinx documentation. For the latter, if you use PyCharm, I have a setup to easily generate Sphinx documentation using [a helper file](https://github.com/mcarans/pycharm-helpers/blob/master/sphinx-apidoc.sh) I've made. 

I've not used pbr before. It looks like as with other similar tools, you create a version in Git and then it takes that version for the package. (it also looks like to it does a bunch of other stuff besides.) Another way to do it that I used is to have a version.txt file in the repo which is used everywhere. I like this approach because there is no magic involved, but of course you have to remember to update it: it depends how much you want to automate. The version.txt file is used by the library itself to [report its version](https://github.com/OCHA-DAP/hdx-python-country/blob/main/src/hdx/location/version.py) which can be useful when debugging. It is also used for the package version (eg. on PyPI) and for the GitHub tag. 

To build a package with the [setup.py](https://github.com/OCHA-DAP/hdx-python-country/blob/main/setup.py)/[setup.cfg](https://github.com/OCHA-DAP/hdx-python-country/blob/main/setup.cfg) I've created just involves doing: `python setup.py clean package`. Note that the `clean` is a customised one as the default setuptools clean unfortunately misses some folders. The package is published to PyPI with `python setup.py publish` and that also creates a tag in GitHub (using the version in version.txt). The various utilities used by setup.py come from [HDX Python Utilities](https://pypi.org/project/hdx-python-utilities/).

As per our call, you could have a src folder, under which aatools (and any other modules) would sit. It can be made to work with pytest using the pytest-pythonpath package and [pytest.ini](https://github.com/OCHA-DAP/hdx-python-country/blob/main/pytest.ini). You would not need to `import src.aatools`, `import aatools` would work even in tests.

For running the tests (and flake8), you can use a GitHub Actions setup like [this](https://github.com/OCHA-DAP/hdx-python-country/blob/main/.github/workflows/run-python-tests.yml). The coverage from the tests is shown in [coveralls.io](https://coveralls.io/) and uses the setup in [coverage.rc](https://github.com/OCHA-DAP/hdx-python-country/blob/main/.coveragerc).

Please let me know if you need more info on any of this.

aatools/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""aa-tools: Toolbox for Anticipatory Action.
+
+The ``aatools`` library contains modules designed to assist with
+downloading and processing the data required for climate- and weather-related
```
Tinkaa, line 4: nitpicking, but wouldn't limit it to climate- and weather-related aa for the future

hannahker:
Not much to add either from the comments already added. This looks really amazing overall and super excited to see this progress! 

Throwing in my agreement about removing the automated changelog. Not sure my commits at least are that clean 😬 

Really love seeing these docs too!

PR #3, opened 2021-10-08
Configuration

turnerm:
A possible implementation of a centralized configuration. Some thoughts:

- Right now I've thrown every configuration in the `config` module itself. This includes a 1-liner class for the `codab` module that has the base directory name. Perhaps this type of thing belongs closer to the module itself?
- On a related note, the COD AB `_get_raw_base_dir` method is something that will likely be repated across modules, maybe  it should be refactored to `config` itself. 
- It would be great to more explicitly validate the country YAML file format
- We should also implement country config file reading from the local directory as well as one set by an environment variable
- The behaviour when a parameter is missing from the config could probably be improved


src/aatoolbox/utils/codab.py
```diff
@@ -0,0 +1,43 @@
...
+    Parameters
+    ----------
+    iso3 : (str)
+        country iso3
+    """
+
+    def __init__(self, iso3: str):
+        super().__init__(iso3=iso3, module_base_dir=MODULE_BASE_DIR)
+
+    def get_admin0(self):
```
Tinkaa, line 22: I know it was more about the config than specific codab stuff. But is there a reason that you decided to do this per admin level? Would it be an idea to instead pass the admin level as parameter? (else we need 5 functions for adm level 0 - 4)

src/aatoolbox/utils/codab.py
```diff
@@ -0,0 +1,43 @@
...
+        --------
+        >>> from aatoolbox.utils.codab import CodAB
+        >>> # Get admin 0 boundaries for Nepal
+        >>> codab = CodAB("npl")
+        >>> npl_admin0 = codab.get_admin0()
+        """
+        shapefile = (
+            self.get_public_raw_dir()
+            / self.config.country.codab.base_dir
+            / f"{self.config.country.codab.base_zip}"
```
Tinkaa, line 40: Thinking out loud, is it too much to require it to be a zip file? At least for most countries I didn't use the zip but unpacked it. Not sure what is the most common approach, thinking about users outside our team

src/aatoolbox/config/country.py
```diff
@@ -0,0 +1,39 @@
...
+from aatoolbox.utils.io import parse_yaml
+
+
+class _CodAdmin(object):
+    def __init__(self, base_layer: str = None, **kwargs):
+        self.base_layer = base_layer
+
+
+class _CodABConfig(object):
+    def __init__(self, base_dir: str = None, base_zip: str = None, **kwargs):
```
Tinkaa, line 13: Would it be useful here to add a comment that the `base_dir` and `base_zip` parameter input to `__init__` are automatically extracted from `object`? Or is that basic programming knowledge that I just didn't have?

src/aatoolbox/utils/codab.py
```diff
@@ -0,0 +1,43 @@
+"""Retrieve COD administrative boundaries."""
```
Tinkaa, line 1: Details, but I would add a bit more explanation / link to what the CODs are

turnerm: Good point! To do for when we expand this class

src/aatoolbox/utils/codab.py
turnerm: I think we should keep it as whatever format it is downloaded as from HDX, which I think is often a zip, as ideally we want to automate the downloading part as well

turnerm: But we will have to refactor this function in the future to accomodate the possible file types

PR #10, opened 2021-10-15
Add keyword-arguments extension for flake8

caldwellst:
Simple addition of [flake8-keyword-arguments](https://pypi.org/project/flake8-keyword-arguments/) for linting as referenced in the [Internal Coding Guidelines doc](https://docs.google.com/document/d/1Ub9zFNqJ-_eV2FDXASJARNhLkWoVGTAa8Y8dHbc53DM/edit?usp=sharing).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,9 @@ repos:
           - flake8-deprecated==1.3
           - flake8-docstrings==1.6.0
           - flake8-implicit-str-concat==0.2.0
+          - flake8-keyword-arguments==0.1.0
           - flake8-print==4.0.0
           - flake8-sfs==0.0.3
-          - flake8-keyword-arguments==0.1.0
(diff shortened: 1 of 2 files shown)


Tinkaa:
Great find indeed! Would agree with Monica to at least set the minimum to 1, as in that case there is not really gonna be ambiguity

caldwellst:
Yeah, agreed about minimum being higher than 0, good idea! Set minimum to 1, but realised there are some instances of 2 arguments functions that don't accept keyword arguments. I added exceptions for them and updated the Google Doc to reflect what we have here. Check what I've done (especially for some of the exceptions as maybe I missed ways to pass keyword arguments?) and merge if okay!

Tinkaa (approved):
Looks good! 
@turnerm would agree that we are over-versioning.. especially with all the changes we gonna make in the near future

PR #16, opened 2021-10-20
Switch markdownlint to Node.js to fix #12

### LINKED ISSUE #12: Changelog markdown linting
The markdown linter doesn't like duplicate headings, which should be an exception for the changelog. However, I can't seem to create an exception for the entire file as described [here](https://github.com/DavidAnson/markdownlint#configuration)

caldwellst:
After a quick investigation again, I finally found the [Node.js markdownlint pre-commit](https://github.com/igorshubovych/markdownlint-cli), which allows full file ignoring.


turnerm (approved):
This is so great, thanks Seth!

PR #19, opened 2021-10-21
Raster module

caldwellst:
Working towards the IRI implementation in the toolbox, the code relied on 3 out of the 4 functions within `raster_manipulation.py` so wanted to start a raster module here. There is still lots of work to do here, but I wanted to put this PR out there now for a bit of review and to get y'alls thoughts on what's missing, things to improve on, things I've missed, or anything else relevant.

I also want to agree on a clear way forward before proceeding. For myself, I think the most natural way would actually to build in a class here and make the module functions already there as methods. This would also be expanded to methods to download the data and otherwise manipulate which the IRI could be slightly refactored to inherit from/use, as could the ARC2 scripts I've worked on, and maybe others. But obviously would be good to agree on a common way forward there.

### DIFF
--- a/src/aatoolbox/utils/raster.py
+++ b/src/aatoolbox/utils/raster.py
@@ -32,7 +32,7 @@ def invert_coordinates(
 
     Parameters
     ----------
-        ds : Union[xr.DataArray, xr.Dataset]
+        ds : Union[xarray.DataArray, xarray.Dataset]
             Dataset with values and coordinates.
         lon_coord : str
             Longitude coordinate in ``ds``.
@@ -41,16 +41,16 @@ def invert_coordinates(
 
     Returns
     -------
-    Union[xr.DataArray, xr.Dataset]
+    Union[xarray.DataArray, xarray.Dataset]
         Dataset or data array with correct coordinate ordering.
 
     Examples
     --------
-    >>> da = xr.DataArray(
-    ...     np.arange(16).reshape(4,4),
-    ...     coords={"lat":np.array([87, 88, 89, 90]),
-    ...             "lon":np.array([70, 69, 68, 67])}
-    ... )
+    >>> da = xarray.DataArray(
+         numpy.arange(16).reshape(4,4),
+         coords={"lat":numpy.array([87, 88, 89, 90]),
+                 "lon":numpy.array([70, 69, 68, 67])}
+    )
     >>> invert_coordinates(ds, "lon", "lat")
     """
     lon_inv, lat_inv = _check_coords_inverted(
@@ -103,23 +103,23 @@ def change_longitude_range(
 
     Parameters
     ----------
-    ds : Union[xr.DataArray, xr.Dataset]
+    ds : Union[xarray.DataArray, xarray.Dataset]
         Dataset with values and coordinates.
     lon_coord : str
         Longitude coordinate in ``ds``.
 
     Returns
     -------
-    Union[xr.DataArray, xr.Dataset]
+    Union[xarray.DataArray, xarray.Dataset]
         Dataset with transformed longitude coordinates.
 
     Examples
     --------
-    >>> da = xr.DataArray(
-    ...     np.arange(16).reshape(4,4),
-    ...     coords={"lat":np.array([87, 88, 89, 90]),
-    ...             "lon":np.array([5, 120, 199, 360])}
-    ... )
+    >>> da = xarray.DataArray(
+         numpy.arange(16).reshape(4,4),
+         coords={"lat":numpy.array([87, 88, 89, 90]),
+                 "lon":numpy.array([5, 120, 199, 360])}
+    )
     >>> change_longitude_range(ds, "lon")
     """
     lon_min = ds.indexes[lon_coord].min()
@@ -147,5 +147,55 @@ def change_longitude_range(
 
     else:
         logger.info("Indeterminate longitude range and no need to convert.")
+
+    return ds
+
+
+def correct_calendar(ds: Union[xr.DataArray, xr.Dataset], time_coord: str):
+    """Correct calendar attribute for recognition by xarray.
+
+    Some datasets come with a wrong calendar attribute that isn't
+    recognized by xarray. This function corrects the coordinate
+    attribute to ensure that a ``calendar`` attribute exists
+    and specifies a calendar alias that is supportable by
+    ``xarray.cftime_range``.
+
+    Currently ensures that calendar attributes that are either
+    specified with ``units="months since"`` or ``calendar="360"``
+    explicitly have ``calendar="360_day"``. If and when further
+    issues are found with calendar attributes, support for
+    conversion will be added here.
+
+    Parameters
+    ----------
+    ds : Union[xarray.DataArray, xarray.Dataset]
+        Dataset with values and coordinates.
+    lon_coord : str
+        Longitude coordinate in ``ds``.
+
+    Returns
+    -------
+    Union[xarray.DataArray, xarray.Dataset]
+        Dataset with transformed calendar coordinate.
+
+    Examples
+    --------
+    >>> da = xarray.DataArray(
+         numpy.arange(64).reshape(4,4,4),
+         coords={"lat":numpy.array([87, 88, 89, 90]),
+                 "lon":numpy.array([5, 120, 199, 360]),
+                 "t":numpy.array([10,11,12,13])}
+    )
+    >>> da["t"].attrs["units"] = "months since 1960-01-01"
+    >>> da = correct_calendar(da, "t")
+    >>> da["t"].attrs["calendar"]
+    """
+    if "calendar" in ds[time_coord].attrs.keys():
+        if ds[time_coord].attrs["calendar"] == "360":
+            ds[time_coord].attrs["calendar"] = "360_day"
+
+    elif "units" in ds[time_coord].attrs.keys():
+        if "months since" in ds[time_coord].attrs["units"]:
+            ds[time_coord].attrs["calendar"] = "360_day"
 
     return ds
--- a/src/aatoolbox/utils/raster.py
+++ b/src/aatoolbox/utils/raster.py
@@ -3,18 +3,19 @@
 import logging
 from typing import Union
 
-import xarray
+import numpy as np
+import xarray as xr
 
 logger = logging.getLogger(__name__)
 
 
 def invert_coordinates(
-    ds: Union[xarray.Dataset, xarray.DataArray],
+    ds: Union[xr.Dataset, xr.DataArray],
     lon_coord: str,
     lat_coord: str,
 ):
     """
-    Invert latitude and longitude in xarray.DataSet.
+    Invert latitude and longitude in ``ds``.
 
     This function checks for inversion of latitude and longitude
     and inverts them if needed. Datasets with inverted coordinates
@@ -31,16 +32,16 @@ def invert_coordinates(
 
     Parameters
     ----------
-        ds : xarray.DataSet | xarray.DataArray
+        ds : Union[xr.DataArray, xr.Dataset]
             Dataset with values and coordinates.
         lon_coord : str
-            Longitude coordinate in `ds`.
+            Longitude coordinate in ``ds``.
         lat_coord : str
-            Latitude coordinate in `ds`.
+            Latitude coordinate in ``ds``.
 
     Returns
     -------
-    xarray.DataSet | xarray.DataArray
+    Union[xr.DataArray, xr.Dataset]
         Dataset or data array with correct coordinate ordering.
 
     Examples
@@ -48,7 +49,7 @@ def invert_coordinates(
     >>> da = xr.DataArray(
     ...     np.arange(16).reshape(4,4),
     ...     coords={"lat":np.array([87, 88, 89, 90]),
-                    "lon":np.array([70, 69, 68, 67])}
+    ...             "lon":np.array([70, 69, 68, 67])}
     ... )
     >>> invert_coordinates(ds, "lon", "lat")
     """
@@ -80,3 +81,71 @@ def _check_coords_inverted(ds, lon_coord, lat_coord):
     lon_start = ds[lon_coord][0].item()
     lon_end = ds[lon_coord][ds.dims[lon_coord] - 1].item()
     return lon_start > lon_end, lat_start < lat_end
+
+
+# TODO: understand when to change longitude range for rasterstats
+# to work and when not!!!
+def change_longitude_range(
+    ds: Union[xr.DataArray, xr.Dataset],
+    lon_coord: str,
+):
+    """Convert longitude range between -180 to 180 and 0 to 360.
+
+    For some raster data, outputs are incorrect if longitude ranges
+    are not converted from 0 to 360 to -180 to 180, such as the IRI
+    seasonal forecast, but *not* the IRI CAMS observational terciles,
+    which are incorrect if ranges are stored as -180 to 180.
+
+    ``change_longitude_range()`` will convert ``ds`` between the
+    two coordinate systems based on its current state. If coordinates
+    lie solely between 0 and 180 then there is no need for conversion
+    and the input ``ds`` will be returned.
+
+    Parameters
+    ----------
+    ds : Union[xr.DataArray, xr.Dataset]
+        Dataset with values and coordinates.
+    lon_coord : str
+        Longitude coordinate in ``ds``.
+
+    Returns
+    -------
+    Union[xr.DataArray, xr.Dataset]
+        Dataset with transformed longitude coordinates.
+
+    Examples
+    --------
+    >>> da = xr.DataArray(
+    ...     np.arange(16).reshape(4,4),
+    ...     coords={"lat":np.array([87, 88, 89, 90]),
+    ...             "lon":np.array([5, 120, 199, 360])}
+    ... )
+    >>> change_longitude_range(ds, "lon")
+    """
+    lon_min = ds.indexes[lon_coord].min()
+    lon_max = ds.indexes[lon_coord].max()
+
+    if lon_max > 180:
+        logger.info("Converting longitude from 0 360 to -180 to 180.")
+
+        ds = ds.assign_coords(
+            {lon_coord: ((ds[lon_coord] + 180) % 360) - 180}
+        ).sortby(lon_coord)
+
+    elif lon_min < 0:
+        logger.info("Converting longitude from -180 to 180 to 0 to 360.")
+
+        ds = ds.assign_coords(
+            {
+                lon_coord: np.where(  # noqa: FKA01
+                    ds[lon_coord] < 0,
+                    ds[lon_coord] + 360,
+                    ds[lon_coord],
+                )
+            }
+        ).sortby(lon_coord)
+
+    else:
+        logger.info("Indeterminate longitude range and no need to convert.")
+
+    return ds
--- a/src/aatoolbox/utils/raster.py
+++ b/src/aatoolbox/utils/raster.py
@@ -1,0 +1,82 @@
+"""Utilities to manipulate and analyze raster data."""
+
+import logging
+from typing import Union
+
+import xarray
+
+logger = logging.getLogger(__name__)
+
+
+def invert_coordinates(
+    ds: Union[xarray.Dataset, xarray.DataArray],
+    lon_coord: str,
+    lat_coord: str,
+):
+    """
+    Invert latitude and longitude in xarray.DataSet.
+
+    This function checks for inversion of latitude and longitude
+    and inverts them if needed. Datasets with inverted coordinates
+    can produce incorrect results in certain functions like
+    ``rasterstats.zonal_stats()``. Correctly ordered coordinates
+    should be:
+
+    * latitude: Largest to smallest.
+    * longitude: Smallest to largest.
+
+    If ``ds`` already has correct coordinate ordering, it is
+    directly returned. Function largely copied from
+    https://github.com/perrygeo/python-rasterstats/issues/218.
+
+    Parameters
+    ----------
+        ds : xarray.DataSet | xarray.DataArray
+            Dataset with values and coordinates.
+        lon_coord : str
+            Longitude coordinate in `ds`.
+        lat_coord : str
+            Latitude coordinate in `ds`.
+
+    Returns
+    -------
+    xarray.DataSet | xarray.DataArray
+        Dataset or data array with correct coordinate ordering.
+
+    Examples
+    --------
+    >>> da = xr.DataArray(
+    ...     np.arange(16).reshape(4,4),
+    ...     coords={"lat":np.array([87, 88, 89, 90]),
+                    "lon":np.array([70, 69, 68, 67])}
+    ... )
+    >>> invert_coordinates(ds, "lon", "lat")
+    """
+    lon_inv, lat_inv = _check_coords_inverted(
+        ds=ds, lon_coord=lon_coord, lat_coord=lat_coord
+    )
+
+    # Flip the raster as necessary (based on the flags)
+    inv_dict = {}
+
+    if lon_inv:
+        logger.info("Longitude was inverted, reversing coordinates.")
+        inv_dict[lon_coord] = ds[lon_coord][::-1]
+
+    if lat_inv:
+        logger.info("Latitude was inverted, reversing coordinates.")
+        inv_dict[lat_coord] = ds[lat_coord][::-1]
+
+    if inv_dict:
+        ds = ds.reindex(inv_dict)
+
+    return ds
+
+
+def _check_coords_inverted(ds, lon_coord, lat_coord):
+    """Check if latitude and longitude inverted."""
+    lat_start = ds[lat_coord][0].item()
+    lat_end = ds[lat_coord][ds.dims[lat_coord] - 1].item()
+    lon_start = ds[lon_coord][0].item()
+    lon_end = ds[lon_coord][ds.dims[lon_coord] - 1].item()
+    return lon_start > lon_end, lat_start < lat_end
--- a/src/aatoolbox/utils/raster.py
+++ b/src/aatoolbox/utils/raster.py
@@ -1,9 +1,12 @@
 """Utilities to manipulate and analyze raster data."""
 
 import logging
-from typing import Union
+from typing import Any, Dict, List, Union
 
+import geopandas as gpd
 import numpy as np
+import pandas as pd
+import rioxarray
 import xarray as xr
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


turnerm:
> Working towards the IRI implementation in the toolbox, the code relied on 3 out of the 4 functions within raster_manipulation.py so wanted to start a raster module here. There is still lots of work to do here, but I wanted to put this PR out there now for a bit of review and to get y'alls thoughts on what's missing, things to improve on, things I've missed, or anything else relevant.

Yes good idea to break it down! When I made all those issues I didn't necessarily intend for them all to be taken care of in one PR (although perhaps I should have split them up more, I was being lazy)

> I also want to agree on a clear way forward before proceeding. For myself, I think the most natural way would actually to build in a class here and make the module functions already there as methods. This would also be expanded to methods to download the data and otherwise manipulate which the IRI could be slightly refactored to inherit from/use, as could the ARC2 scripts I've worked on, and maybe others. But obviously would be good to agree on a common way forward there.

Do you mean making a class that contains all the raster functions? I would maybe keep them as is and then call them in with the (future) IRI module as needed. A few reasons for this:
- I would argue that they don't really belong as part of the IRI class, as they don't really care about the state (they only need the dataset, they would just be static methods)
- Similarly, classes should follow the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)
- On their own they don't really constitute an object, which should have both a state and behaviour

Given the above I should probably clarify my intentions with the `AAToolboxBase` class (you can see how I used it in `codab.py`). The idea behind that is to have an object that essentially represents the DAG vertices, with an initializer that sets the state based on the country ISO3, and then contains a few convenience methods for finding directories (and anything else that might be universal). So I would propose to subclass that for the IRI download / processing, and then call the above raster methods on the data when you need them.

I want to add that I think the most important thing with these data manipulation functions if for us to add unit tests. 

[Also, final caveat one what I said above about the base class. If you look at the GloFAS class in the old AA repo, it is actually two DAG vertices (download & process). I'm wondering if it would be better to split those up or not. We also need to think about it terms of API usability].

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,324 @@
...
+    """Check if latitude and longitude inverted."""
+    lat_start = ds[lat_coord][0].item()
+    lat_end = ds[lat_coord][ds.dims[lat_coord] - 1].item()
+    lon_start = ds[lon_coord][0].item()
+    lon_end = ds[lon_coord][ds.dims[lon_coord] - 1].item()
+    return lon_start > lon_end, lat_start < lat_end
+
+
+# TODO: understand when to change longitude range for rasterstats
+# to work and when not!!!
```
Tinkaa, line 90: Haha I like exclamation marks in my old comments. But think this can be removed now. As far as I understand rasterstats overlays the array, and incorporates the CRS but this isn't able to account for an incorrect range. This is one of the main reasons we want to rely more on the rioxarray functions for computing stats than rasterstats. At the same time, if we know the range is not in common conventions, we prefer to change it just to be sure

caldwellst: In the end, I have removed the doctest implementation until we can sort out #23.

caldwellst:
Alright, there is more to do (in terms of implementing functionality from `pa-anticipatory-action`, but now it is ready for review in terms of the method of implementing. I've put in a mixin class that will carry methods applicable to both datasets and data arrays, and the data array extension class. The dataset class will essentially be the same thing (although a lot of looping through component data arrays and using the methods defined now). Before starting that, It think would be good to get agreement that this approach is the best way forward.

Look forward to your reviews @turnerm @Tinkaa!

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+    # setting attributes to avoid mypy error, from SO
+    # https://stackoverflow.com/questions/53120262/mypy-how-to-
+    # ignore-missing-attribute-errors-in-mixins/53228204#53228204
+    _get_obj: Callable
+
+    def __init__(self, xarray_obj):
+        super().__init__(xarray_obj)
+
+        # Adding lat/lon to set of default spatial dims
+        if "lat" in self._obj.dims and "lon" in self._obj.dims:
```
turnerm, line 53: So useful!

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+import logging
+from typing import Any, Callable, Dict, List, Union
+
+import geopandas as gpd
+import numpy as np
+import pandas
+import pandas as pd
+import rioxarray
+import xarray
+import xarray as xr
```
turnerm, line 34: You should only import xarray once

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+            return self._t_dim
+        raise rioxarray.exceptions.DimensionError(
+            "Time dimension not found. 'aat.set_time_dim()' or "
+            "using 'rename()' to change the dimension name to "
+            f"'t' can address this.{_get_data_var_message(self._obj)}"
+        )
+
+    def set_time_dim(
+        self, t_dim: str, inplace: bool = False
+    ) -> Union[xarray.DataArray, xarray.DataSet]:
```
turnerm, line 77: Make sure to use `Dataset` instead of `DataSet` (it's in a few more places as well)

turnerm (commented):
Shoot, I've totally run out of time today so I will just leave it here for now. This looks amazing and I'm keen to keep poking around, the only issue I'm having is getting xarray to register the accessor. To be honest I've always been a bit mystified by how rioxarray works -- you import rioxarray and suddenly your xarrays have this `rio` property. I've tried importing `raster` but the same magic didn't happen with `aat`. I looked a bit at the e.g. `__init__.py` of rioxarray but haven't found anything obvious. Something for next week I guess!

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+        # Adding lat/lon to set of default spatial dims
+        if "lat" in self._obj.dims and "lon" in self._obj.dims:
+            self._x_dim = "lon"
+            self._y_dim = "lat"
+
+        # Managing time coordinate default dims
+        self._t_dim = None
+        for t in ["t", "T"]:
+            if t in self._obj.dims:
+                self._t_dim = t
```
Tinkaa, line 61: would "time" aslo be a default? Or is that already added by standard (rio)xarray

Tinkaa: hmm I see later on the `set_time_dim()` where you use `time` as an example. Maybe I am missing the logic, but why is `time` not in the standard set (as I have seen that being used quite often)?

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+
+    def change_longitude_range(
+        self, inplace: bool = False
+    ) -> xarray.DataArray:
+        """Convert longitude range between -180 to 180 and 0 to 360.
+
+        For some raster data, outputs are incorrect if longitude ranges
+        are not converted from 0 to 360 to -180 to 180, such as the IRI
+        seasonal forecast, but *not* the IRI CAMS observational terciles,
+        which are incorrect if ranges are stored as -180 to 180.
```
Tinkaa, line 271: Honestly this CAMS thing, I would have to double-check as I was doing that in the first weeks I learned about geospatial processing. Haven't used the functionality structurally for anything else than IRI. And it is more standard to use -180 to 180 so I am doubting if we should even provide functionality to convert to 0-360 or that might cause incorrect use. 

I will add to my to do to look into it a bit

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+                        data_obj[self.x_dim] < 0,
+                        data_obj[self.x_dim] + 360,
+                        data_obj[self.x_dim],
+                    )
+                }
+            ).sortby(self.x_dim)
+
+        else:
+            logger.info(
+                "Indeterminate longitude range and no need to convert."
```
Tinkaa, line 327: I think this is not fully correct though.. Cause if all values are between 0 and 180, you don't know if the full range the coordinates are in -180 to 180 or 0 to 360. So maybe it is now in -180 to 180 and to shift it to 0-360 we would have to add 180.. 
It could be solved by giving as input arg the current coordinate range? But maybe also overcomplicating

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+        feature_col : str
+            Column in ``gdf`` to use as row/feature identifier.
+        stats_list : List[str], optional
+            List of statistics to calculate, by default None.
+            Passed to ``get_attr()``.
+        percentile_list : List[int], optional
+            List of percentiles to compute, by default None.
+        all_touched : bool, optional
+            If ``True`` all cells touching the region will be
+            included, by default False.
```
Tinkaa, line 360: I know I wrote this but would add if ``False`` only cells with their centre in the region will be included

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+        df_list = []
+
+        if stats_list is None:
+            stats_list = ["mean", "std", "min", "max", "sum", "count"]
+
+        for feature in gdf[feature_col].unique():
+            gdf_adm = gdf[gdf[feature_col] == feature]
+
+            da_clip = data_obj.rio.set_spatial_dims(
+                x_dim=self.x_dim, y_dim=self.y_dim
```
Tinkaa, line 384: oehh I like this

Tinkaa (commented):
Looks great @caldwellst , learned many new things :) 
Few small comments/questions.

For the extension to datasets, do you think it can inherit from `AatRasterArray` or we would have do duplicate all the code? 

Don't know a better name for the accessor either but will think about it. 

Out of scope of this PR: I don't know how deep you dived into xarray vs rioxarray but if you got interesting learnings from that, we could dot them down?

src/aatoolbox/utils/raster.py
caldwellst: So, interestingly, I think thiswas being pushed back in there during pre-commit, and I hadn't noticed. This is because I had set the return typing to be `xarray.DataArray` rather than `xr.DataArray`, which helped automatically generate the docstring without the `xr` alias.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
+"""
```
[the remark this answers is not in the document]
caldwellst: Oh man, genius find!! Definitely will add and use.

src/aatoolbox/utils/raster.py
caldwellst: Okay thanks @Tinkaa, will add as a separate issue to ensure proper tracking. Will leave in for now!

src/aatoolbox/utils/raster.py
caldwellst: @Tinkaa Conversions between -180 to 180 and 0 to 360 only convert values that are `< 0` in the first instance or `> 180` in the second. Values between `0` and `180` are identical in both coordinate systems, hence why no shift would need to happen. I had to look this up and confirm myself while thinking through the implemented method ha. [This website](http://gis.humboldt.edu/OLM/Courses/old/GSP_101/01_IntroAndReivew/Presentation/Latitude_and_Longitude.html) helped me wrap my around it.

src/aatoolbox/utils/raster.py
caldwellst: There is definitely no good logic to it and happy to increase the default set as necessary! Was just putting in what I'd commonly seen. Have now added `time` and if you have any other defaults you think are good, let me know @Tinkaa @turnerm. I hadn't seen `time` before but had also seen `F` out there, but thought that might not be an appropriate default and we should just require manually specifying in the `datasource`?

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AatRasterMixin:
+    """AA toolbox mixin base class."""
+
+    # setting attributes to avoid mypy error, from SO
+    # https://stackoverflow.com/questions/53120262/mypy-how-to-
+    # ignore-missing-attribute-errors-in-mixins/53228204#53228204
+    _get_obj: Callable
```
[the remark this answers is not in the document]
caldwellst: That's a good point. I have no idea actually. Seems a bit strange/nonsensical for one to be caught but another not to be. Could it be because if `_obj` doesn't exist it will default to `None` while there's no default if a function/Callable doesn't exist?

src/aatoolbox/datasources/cds/area.py
```diff
@@ -126,7 +126,7 @@ class AreaFromStations(Area):
     Examples
     --------
     >>> stations = {"station1": Station(lon=1.0, lat=2.0),
-    >>>             "station2": Station(lon=3.0, lat=4.0)}
+    ...             "station2": Station(lon=3.0, lat=4.0)}
```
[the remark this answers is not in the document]
caldwellst: God, trying to find something coherent and useful, but really struggling here @turnerm. This website has some basics of the actual [doctest framework](https://adhikary.net/en/2018/06/20/effortless-doctests-in-python/), which you can also find on the [doctest documentation](https://docs.python.org/3/library/doctest.html). However, can't find some clear and easy cheat sheet! The first link however made it most clear, which is basically that it tries to emulate Python Shell (couldn't find a "cheatsheet" for that), which is why we use the ellipsis to denote additional lines after the first but prior to output.

tests/test_raster.py
```diff
@@ -0,0 +1,11 @@
+"""Tests for the raster utilities module."""
+import doctest
+import unittest  # noqa: F401
+
+from aatoolbox.utils import raster
+
+
+def load_tests(loader, tests, ignore):
```
turnerm, line 8: I get the same coverage when I remove this function, do you find the same @caldwellst? 

I guess my one concern with the doctests is that they don't actually assert the method output. For example if I change `invert_coords` to not invert the coordinates at all, then this won't be picked up by the doc tests (I think).

turnerm (commented):
Part 2 of my review done, sorry for the delay! Look great though and can't wait to see the updates.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,447 @@
...
+        ...          "t":numpy.array([10,11,12,13])}
+        ... )
+        >>> da["t"].attrs["units"] = "months since 1960-01-01"
+        >>> da_crct = da.aat.correct_calendar()
+        >>> da_crct["t"].attrs["calendar"]
+        '360_day'
+        """
+        data_obj = self._get_obj(inplace=inplace)
+        if "calendar" in data_obj[self.t_dim].attrs.keys():
+            if data_obj[self.t_dim].attrs["calendar"] == "360":
```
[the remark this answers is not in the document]
caldwellst: Hmm, so, putting on the same line breaks the formatting rules so then they are auto-formatted with a line break. I think reads better in this manner because of that, although otherwise would put on the same line. What do you think @turnerm? And yeah, it's intentional if second is false, no changes made. Only want to change `360` to `360_day` since `360` isn't a valid calendar recognised by NetCDF (neither are months since). Here's a reference [GitHub issue](https://github.com/Unidata/netcdf4-python/issues/434) that I've now added to the documentation of the function.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
+"""
+Utilities to manipulate and analyze raster data.
+
+This module extends xarray and rioxarray to provide
+additional functionality for raster processing and
+post-processing. The extension is based on the
+guidance for how to extend xarray:
+
+http://xarray.pydata.org/en/stable/internals/extending-xarray.html
```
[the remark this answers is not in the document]
caldwellst: Yes I should, I actually didn't know how so left like this ha!!

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+        ...  coords={"lat":numpy.array([90, 89, 88, 87]),
+        ...          "lon":numpy.array([70, 69, 68, 67])}
+        ... )
+        >>> da.aat._check_coords_inverted()
+        (True, False)
+        """
+        data_obj = self._get_obj(inplace=False)
+        lat = data_obj.get_index(self.y_dim)
+        lat_start = lat[0]
+        lat_end = lat[-1]
```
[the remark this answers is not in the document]
caldwellst: **Way clearer!!**

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+        Examples
+        --------
+        >>> import xarray
+        >>> import numpy
+        >>> da = xarray.DataArray(
+        ...  numpy.arange(16).reshape(4,4),
+        ...  coords={"lat":numpy.array([90, 89, 88, 87]),
+        ...          "lon":numpy.array([70, 69, 68, 67])}
+        ... )
+        >>> da.aat._check_coords_inverted()
```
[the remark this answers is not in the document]
caldwellst: Okay fair enough. I always try to be as explicit as possible, just so it's easier if we ever come back to specific methods but yeah, maybe not worth the effort. I won't remove but will keep in mind for the future.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+        lat_end = lat[-1]
+        lon = data_obj.get_index(self.x_dim)
+        lon_start = lon[0]
+        lon_end = lon[-1]
+        return lon_start > lon_end, lat_start < lat_end
+
+    def change_longitude_range(self, inplace: bool = False) -> xr.DataArray:
+        """Convert longitude range between -180 to 180 and 0 to 360.
+
+        For some raster data, outputs are incorrect if longitude ranges
```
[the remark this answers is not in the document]
caldwellst: No, I haven't used in practice so would need @Tinkaa's input.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+        >>> import xarray
+        >>> import numpy
+        >>> da = xarray.DataArray(
+        ...  numpy.arange(16).reshape(4,4),
+        ...  coords={"lat":numpy.array([87, 88, 89, 90]),
+        ...          "lon":numpy.array([5, 120, 199, 360])}
+        ... )
+        >>> da_inv = da.aat.change_longitude_range()
+        >>> da_inv.get_index("lon")
+        Int64Index([-161, 0, 5, 120], dtype='int64', name='lon')
```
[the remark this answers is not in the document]
caldwellst: Great thanks! Also revealed an issue with my `inplace` argument that I've fixed.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+            data_obj = data_obj.assign_coords(
+                {self.x_dim: ((data_obj[self.x_dim] + 180) % 360) - 180}
+            ).sortby(self.x_dim)
+
+        elif lon_min < 0:
+            logger.info("Converting longitude from -180 to 180 to 0 to 360.")
+
+            data_obj = data_obj.assign_coords(
+                {
+                    self.x_dim: np.where(  # noqa: FKA01
```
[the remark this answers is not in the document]
caldwellst: Yeah, I looked into this and couldn't find something mathematically consistent. `(x - 180) % 360 + 180` works great for the negative numbers (or 0), but those in the range from 1 to 180 are then `> 360`. Essentially all methods I remember looking at I was struggling to find a statement that left the positive numbers in place.

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
turnerm: I would tend towards using an `and`, even if it results in line breaks, because as a rule of thumb you want to reduce nesting (e.g. [flat is better than nested](https://www.python.org/dev/peps/pep-0020/)). 

But my question was really about the `elif` following: Do you mean for it to be `elif` or just an `if`?

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
caldwellst: Any ideas on an equation that might work or should I leave it @turnerm?

tests/test_raster.py
caldwellst: Son of a gun. Took me a long while to figure out, but clearly the doctests weren't running, because they do assert the method output (test by just right clicking in PyCharm and running). But, my solution to using doctest through unittest [doesn't work](https://docs.pytest.org/en/6.2.x/unittest.html).

> Up to this point pytest does not have support for the following features:
> - load_tests protocol
> - subtests

So, essentially, the `--with-doctests` argument passed to `pytest` is the recommended way, but gives us inflated coverage across non-tested modules. The `unittest` recommended way of defining `load_tests()` is actually not supported by `pytest`. So, I [found a simple way](https://stackoverflow.com/questions/5681330/using-doctests-from-within-unittests) to manually load those tests with `unittest` and it all works now!!

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
turnerm: Oops, I shouldn't math so close to my bedtime, thanks for checking (I thought I had). Am I crazy or does just `x % 360` work?

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
caldwellst: Ah okay! Will add in. Regarding the `elif`, yeah, would keep that way since the change in the end is the same. I thought about putting both changes on the same line of code, but based on your instructions to use `logger`, I think actually is better to keep separate so the logging information is clearer (one is correcting an incorrect attribute, one is adding an attribute that is equivalent). Let me know what you think.

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
caldwellst: GENIUS!

caldwellst:
Alright @turnerm @Tinkaa, putting this in for final review. I am working on the final piece which is testing for the `compute_raster_stats()` methods, the rest I've finally gotten working through doctest. Beyond that, we just have the unresolved conversations above to sort out. I have not used `docrep` for the dataset method of `compute_raster_stats()`, instead am just linking to the array method for now, but `docrep` could be a future implementation. Couldn't get it working, however, so leaving as is for the moment and creating a `docrep` issue which we can use later to implement some functionality like that.

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
caldwellst: Resolving this and referencing in #32 for longer term discussion.

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,443 @@
...
+                data_obj[self.t_dim].attrs["calendar"] = "360_day"
+
+        elif "units" in data_obj[self.t_dim].attrs.keys():
+            if "months since" in data_obj[self.t_dim].attrs["units"]:
+                data_obj[self.t_dim].attrs["calendar"] = "360_day"
+
+        return data_obj if not inplace else None
+
+
+@xr.register_dataarray_accessor("aat")
```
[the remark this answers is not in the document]
Tinkaa: `aat` almost sounds like `aight` when you pronounce it haha but lets not use that. 
I actually think `aat` is fine.

src/aatoolbox/utils/raster.py
Tinkaa: Yes definitely wouldn't add `F`. Only have seen that used by IRI related to forecasted times. But saw you added `time` which seems good

src/aatoolbox/utils/raster.py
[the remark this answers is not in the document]
caldwellst: At some point, mypy called out `_obj` as well, so also added there. Strange.

src/aatoolbox/utils/raster.py
```diff
@@ -445,3 +462,107 @@ def compute_raster_statistics(
 
         df_zonal_stats = pd.concat(df_list).reset_index(drop=True)
         return df_zonal_stats
+
+
+@xr.register_dataset_accessor("aat")
+class AatRasterDataset(AatRasterMixin, RasterDataset):
```
Tinkaa, line 468: class looks good! Didn't know `ds["var"]` makes you loose attributes..

Tinkaa:
Ready to merge imo (after resolving the conflicts)

src/aatoolbox/utils/raster.py
caldwellst: Since the classes now rely on a set `x_dim` and `y_dim` rather than functions just specifying those explicitly, it is now a problem, but don't think it was previously (since we didn't set those I think, at least in code I've seen).

tests/test_raster.py
```diff
@@ -0,0 +1,11 @@
+"""Tests for the raster utilities module."""
+import doctest
+import unittest  # noqa: F401
```
turnerm, line 3: I think you can take out the noqa now

turnerm: I think you can take out the noqa

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,568 @@
...
+            obj.rio.set_crs(self._crs, inplace=True)
+
+        # raster module attributes
+        if self._t_dim is not None:
+            obj.aat.set_time_dim(self._t_dim, inplace=True)
+
+        return obj
+
+    def compute_raster_stats(
+        self, vars: Union[List[str], None] = None, **kwargs: Any
```
turnerm, line 543: You shouldn't use the name `vars` because it's one of the [builtin functions](https://docs.python.org/3/library/functions.html)

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,568 @@
...
+            Keyword arguments passed to the array method
+            :meth:`~AatRasterArray.compute_raster_stats`
+
+        Returns
+        -------
+        List[pandas.DataFrame]
+            List of raster statistics data frames.
+        """
+        vars = self.vars if vars is None else vars
+        stats = [self._obj[var].compute_raster_stats(**kwargs) for var in vars]
```
turnerm, line 567: You could also consider returning a dictionary with `vars` as the key

src/aatoolbox/utils/raster.py
```diff
@@ -0,0 +1,568 @@
...
+
+
+@xr.register_dataset_accessor("aat")
+class AatRasterDataset(AatRasterMixin, RasterDataset):
+    """AA toolbox extension for xarray.Dataset."""
+
+    def __init__(self, xarray_object):
+        super().__init__(xarray_object)
+
+    def get_raster_array(self, var: str) -> xr.DataArray:
```
turnerm, line 474: Annoying that you have to write this method, but good catch. Does `rioxarray` try to handle this at all?

turnerm (approved):
Approved! Just a few last small points.

src/aatoolbox/utils/raster.py
caldwellst: Oh damn good catch!!

src/aatoolbox/utils/raster.py
caldwellst: From looking through [their code](https://github.com/corteva/rioxarray/blob/master/rioxarray/raster_dataset.py), I didn't see any attempt to deal with it.

PR #21, opened 2021-11-01
Feature/ecmwf seas download

Tinkaa:
First version of how the ecmwf api can be used to download data. 

Mainly a draft cause I was moving around the area functionalities. @turnerm I now left the `AreaFromStations` in `cds/area.py`. But actually what I think it does is create a bounding box around a list of lon-lat coords. I haven't been working with it so didn't want to change it. But would it make sense to rename it to `AreaFromList` or so, and then add it to `utils/area.py`? In that case I don't think we need the `cds/area.py` file anymore as it would only contain one function.  

As always I still have to create tests, but to me that feels it makes more sense after the code has been finalized more. 

Also, I am struggling quite a bit with where to define folder+file names. I imagine part of this could be solved when using classes, but not 100% how, so if you got tips on that, happy to hear!


Tinkaa:
Close in favour of #37

PR #25, opened 2021-11-08
Implement COD AB class

### LINKED ISSUE #9: COD AB manipulation tools
Since we use the COD ABs a lot, it would be useful to have an easy way to download & manipulate them. A class is already started in `utils` but should probably be moved to `datasources` and expanded.

turnerm:
This PR closes #9 and implements the following:
- a `CodAB` class that downloads and returns COD AB as a geopandas dataframe
  - it subclasses the `DataSource` class, which so far mainly provides convenience filepath methods
  - uses a new HDX API caller
- config files for Bangladesh, Ethiopia, Malawi and Nepal with admin boundary info filled in
- a top-level `Pipeline` class that is meant to be the API for (most) users
  - this is meant to serve as a wrapper around the various `DataSources`, passing the relevant parameters from `config`
 - per @caldwellst's suggestion, the `DataSource` classes don't deal with the config objects directly, and the `Pipeline` class provides a bridge between them
 - the `DataSource` subclasses are still intended to have documented public methods for power users
 - the `io` module has a new decorator which can be used to perform caching without code repetition

Questions:
- are there other specific countries that I should add?

For a future PR:
- a processing step that homogenizes column names and saves as a geopackage


src/aatoolbox/config/pathconfig.py
```diff
@@ -0,0 +1,22 @@
...
+@dataclass
+class PathConfig:
+    """Global directory parameters."""
+
+    public: str = "public"
+    private: str = "private"
+    raw: str = "raw"
+    processed: str = "processed"
+
+    def __init__(self, base_dir_env: str = BASE_DIR_ENV):
```
caldwellst, line 19: So, had to do some reading about the `@dataclass` decorator as I'm not familiar, looks very cool. The only thing I notice is it seems it's recommended to use the `__post_init__` method, which becomes the last thing called by `__init__`. In this case, could you take advantage of the data class by changing this to:

```python
@dataclass
 class PathConfig:
     """Global directory parameters."""

     public: str = "public"
     private: str = "private"
     raw: str = "raw"
     processed: str = "processed"
     base_dir_env: str = BASE_DIR_ENV

     def __post_init__(self):
          self.base_path = Path(os.environ[self.base_dir_env])
```

src/aatoolbox/config/pathconfig.py
```diff
@@ -0,0 +1,22 @@
+"""Path configuration."""
+
+import os
+from dataclasses import dataclass
+from pathlib import Path
+
+BASE_DIR_ENV = "AA_DATA_DIR"
```
caldwellst, line 7: How does this work? Is it set here like this because I am able to manually change this in a session or script?

```python
import aatoolbox.config.pathconfig

BASE_DIR_ENV = "OTHER_ENV_VAR"
```

And then I can init a `PathConfig()` with my `"OTHER_ENV_VAR"` as the default? Is there a reason not to just pass in the name of the environment variable directly into the initialisation, and the user just has to specify their environment variable? 

Or is this variable being set just for cleanliness of code in the module and if we change the name of the env var from `"AA_DATA_DIR"` then it's easier for us to just change at the top of the script? 

Sorry, think I'm not getting exactly how this works, so interested to see your thoughts!

src/aatoolbox/datasources/codab.py
```diff
@@ -0,0 +1,76 @@
+"""Retrieve COD administrative boundaries."""
+from pathlib import Path
+
+import geopandas as gpd
+
+from aatoolbox.datasources.datasource import DataSource, check_file_existence
+from aatoolbox.utils.hdx_api import get_dataset_from_hdx
+
+MODULE_BASE = "cod_ab"
```
caldwellst, line 10: Similar question here as above, is it best practice to assign `MODULE_BASE` external to the class? Are users supposed to be able to change it? Or is this just again because it's easier for us to change `MODULE_BASE` at the top of the script if we decide to change the folder name in the future or something?

src/aatoolbox/utils/hdx_api.py
```diff
@@ -0,0 +1,52 @@
...
+    for resource in resources:
+        if resource["name"] == hdx_dataset_name:
+            logger.info(f"Downloading dataset {hdx_dataset_name}")
+            with tempfile.TemporaryDirectory() as tempdir:
+                _, downloaded_filepath = resource.download(folder=tempdir)
+                output_filepath.parent.mkdir(parents=True, exist_ok=True)
+                shutil.copy(downloaded_filepath, output_filepath)
+            logger.info(f"Saved to {output_filepath}")
+            return Path(output_filepath)
+    raise FileNotFoundError(
```
caldwellst, line 49: Love this error catching.

src/aatoolbox/datasources/datasource.py
```diff
@@ -0,0 +1,98 @@
...
+            / self._iso3
+            / self._module_base_dir
+        )
+
+
+# For typing the decorator
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def check_file_existence(filepath_attribute_name: str) -> Callable[[F], F]:
```
caldwellst: Love this.

caldwellst (commented):
Overall looks great to me. Just have a few comments/questions included in there!

src/aatoolbox/pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+            ]  # type: ignore
+        except (IndexError, TypeError):
+            raise AttributeError(
+                f"{custom_layer_number}th custom layer requested but not "
+                f"available in {self._config.iso3.upper()} config file"
+            )
+        return self._get_codab(layer_name=layer_name)
+
+    def _get_codab(self, layer_name):
+        self._codab.download(
```
Tinkaa, line 89: do we want to pass the `cobber` arg as well? might be e.g. that boundaries get updated but you don't want to download the updated boundaries. But maybe that is overthinking it

src/aatoolbox/datasources/codab.py
```diff
@@ -0,0 +1,88 @@
+"""Retrieve COD administrative boundaries."""
```
Tinkaa, line 1: Very general comment, but do we want to have a folder per datasource if num files>=2 and else just have a single `.py` file, or always encapsulate it by a dir?

src/aatoolbox/config/countries/bgd.yaml
```diff
@@ -0,0 +1,6 @@
+iso3: bgd
+codab:
+  hdx_address: administrative-boundaries-of-bangladesh-as-of-2015
```
Tinkaa, line 3: not specific for this PR but just making a note that the address and name might change at some point. Could we see with the HDX team if we can get an update on that or smth?

Tinkaa (commented):
So much cool stuff @turnerm ! Some comments, though some are also not PR specific. 
For country configs, I think we can add others later once we start using them. 

Now just gotta get to using it to experience if the structure works :)

src/aatoolbox/config/countries/bgd.yaml
```diff
@@ -0,0 +1,14 @@
...
+  hdx_dataset_name: bgd_adm_bbs_20201113_SHP.zip
+  admin0:
```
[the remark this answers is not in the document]
turnerm: I actually restructured a bit since you left this comment. What do you think of the new format?

src/aatoolbox/config/pathconfig.py
turnerm: As usual your Googling skills reign supreme. Nice suggestion, fixed!

src/aatoolbox/config/pathconfig.py
turnerm: I set it there because I thought it makes sense (at least for now) as a module-level [constant](https://www.python.org/dev/peps/pep-0008/#constants)

src/aatoolbox/datasources/codab.py
turnerm: Same as before, it's the [PEP 8 convention](https://www.python.org/dev/peps/pep-0008/#constants) for constants. Although in this case I'm not super happy with the name. Basically it's like the base name for the directories and files.

src/aatoolbox/utils/hdx_api.py
turnerm: Aha thanks!

src/aatoolbox/datasources/codab.py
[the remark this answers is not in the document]
turnerm: Ah oops I remember you commented this last time and I forgot to do it. Done now!

src/aatoolbox/config/pathconfig.py
caldwellst: Clear. Since it's intended to be non-public, should we put leading underscores in front of it as described [here](https://www.python.org/dev/peps/pep-0008/#global-variable-names)? I just came across that actually on something I was reviewing of Tinka's!

src/aatoolbox/datasources/codab.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Parameters
+    ----------
+    iso3 : (str)
+        country iso3
+    """
+
+    def __init__(self, iso3: str):
+        super().__init__(
+            iso3=iso3, module_base_dir=MODULE_BASENAME, is_public=True
```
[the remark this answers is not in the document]
turnerm: Yeah, that's a good point and AFG is a tricky case. I think we would make it an optional arg in the config file and then end up passed to `__init__` as an arg, like you suggest. But I would say let's not cross that bridge until we come to it ;)

src/aatoolbox/datasources/codab.py
```diff
@@ -0,0 +1,88 @@
...
+        Returns
+        -------
+        geopandas dataframe with COD AB admin information
+
+        Examples
+        --------
+        >>> from aatoolbox.datasources.codab import CodAB
+        >>> # Get admin 0 boundaries for Nepal
+        >>> codab = CodAB("npl")
+        >>> npl_admin0 = codab.get_admin_layer()
```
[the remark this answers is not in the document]
turnerm: Oops, good catch! We need Seth's doc testser.

src/aatoolbox/datasources/datasource.py
```diff
@@ -0,0 +1,51 @@
...
+        self._path_config = PathConfig()
+        self._raw_base_dir = self._get_base_dir(
+            is_public=is_public, is_raw=True
+        )
+        self._processed_base_dir = self._get_base_dir(
+            is_public=is_public, is_raw=False
+        )
+
+    def _get_base_dir(self, is_public: bool, is_raw: bool) -> Path:
+        public_dir = (
```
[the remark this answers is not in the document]
turnerm: I deserve to be nitpicked, and also I love it so don't worry. 

I used `permission_dir` and `state_dir`, what do you think?

src/aatoolbox/config/countries/bgd.yaml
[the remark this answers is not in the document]
caldwellst: Even better, nice! 👍

src/aatoolbox/config/pathconfig.py
turnerm: That's a good point. We should really setup `__all__` in one of the init's but I will add the underscores for now!

src/aatoolbox/pipeline.py
turnerm: No you're right I was just being lazy. Btw are you okay with using `clobber` instead of `use_cache`? I though it's more clear and universal.

src/aatoolbox/datasources/codab.py
turnerm: That's kind of what I was thinking, because I feel like it's overkill to have packages with just a single module. At the same time, maybe it's inconsistent to have some datasource modules be just within the datasource package while others have their own package. Maybe I'm starting to lean towards directories for everything. What about you?

src/aatoolbox/config/countries/bgd.yaml
turnerm: Yeah that's tricky, for this part of the pipeline we could have it setup to run once a day or something which would serve as a check

src/aatoolbox/datasources/codab.py
Tinkaa: I would say directories. Feels like if we have both directories and single files, it becomes less organized and harder to understand

src/aatoolbox/config/countries/bgd.yaml
Tinkaa: Yes that could work. Also think it is good to ask HDX team what they would recommend. Lets leave it for now and make an issue?

src/aatoolbox/pipeline.py
Tinkaa: I had to google what clobber means haha but fine by me

src/aatoolbox/datasources/codab.py
caldwellst: I would also vote directories here. This was something I was just thinking about where we probably want to have a datasource utils directory (not sure if we want to explicitly call it that in case it confuses with the main utils directory). But things like the Area.py used in CDS is going to be useful across multiple APIs, and we would probably want most of that functionality if not all in a utils directory accessible to others.

src/aatoolbox/datasources/codab.py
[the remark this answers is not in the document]
Tinkaa: Agreed. With the current code that calls the HDX API it wouldn't be able to download files that are on HDX connect right? If so, then lets leave it

Tinkaa (approved):

caldwellst (approved):

src/aatoolbox/datasources/codab.py
turnerm: It depends which packages will use `Area`. If it's just CDS then it should stay there, but otherwise it could maybe go to `utils`? At any rate agreed, I will move COD AB to its own directory!

turnerm: Also that way the `__init__` file of each data source is a good place to explain what exactly the datasource is

src/aatoolbox/config/countries/bgd.yaml
turnerm: Created #30

src/aatoolbox/datasources/codab.py
Tinkaa: Great. Specific to `Area`, I also wanted to use it for another datasource so moved it to `utils` in #21

PR #27, opened 2021-11-09
retrieve files from icpac's ftp server

Tinkaa:
- Functionality to acces ICPAC's FTP server and download a file
- Proposal on how we could use an `.env.example` with `.dotenv`

Leaving at as a draft because
- Would like to hear your opinion on the usage of `.env`
- Question in `retrieve_file_ftp` if you know a better method to check for not None and make mypy happy?


caldwellst (commented):
Put some comments in the code, it looks good! My main other thoughts are:

- The functions should clearly document the environment variables expected for passing in. (and I think both functions should expect env variables, not mixing up passed in arguments vs env variables). That documentation should probably link to more extensive documentation about ICPAC and how to access/setup these environment variables.
- We should also just have a library-wide documentation on env variables, which I think should encourage use of `.env` files but have checking that only ensures the variables are set, not checking how (can we ensure there's no error if a `.env` file doesn't exist)

src/aatoolbox/datasources/icpac/icpac.py
```diff
@@ -0,0 +1,93 @@
...
+        path to save the file to
+
+    Examples
+    --------
+    >>> retrieve_file_ftp(ftp_filepath=
+    ... '/SharedData/gcm/seasonal/202101/' \
+    ... 'PredictedRainfallProbbability-FMA2021_Jan2021.nc',
+    ... output_filepath='example.nc')
+    """
+    load_dotenv()
```
caldwellst, line 76: Should we be using `find_dotenv()` as well here as per [this SO post](https://stackoverflow.com/questions/41546883/what-is-the-use-of-python-dotenv)? Where does `load_dotenv()` on its own expect to find the file?

caldwellst: Reading the documentation of [`load_dotenv()`](https://github.com/theskumar/python-dotenv/blob/45848bb780c26ef0adf7898656f7d3d3f4e2d8ae/src/dotenv/main.py), there is no warning if the file isn't found by default, so might want to be careful here. However, by default, the code also shows that `find_dotenv()` is used, so hopefully shouldn't be too problematic?

src/aatoolbox/datasources/icpac/icpac.py
```diff
@@ -0,0 +1,93 @@
...
+    ftp_address = os.getenv("ICPAC_FTP_ADDRESS")
+    ftp_username = os.getenv("ICPAC_FTP_USERNAME")
+    ftp_password = os.getenv("ICPAC_FTP_PASSWORD")
+    # question: is this the best method to set the vars and raise the error?
+    if None in (ftp_address, ftp_username, ftp_password):
+        raise RuntimeError("One of the ftp variables is not set")
+    # TODO: ugly, is there a better method? if not doing, mypy complains
+    assert ftp_address is not None, ftp_address
```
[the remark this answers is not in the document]
turnerm: Oh so I actually stopped running mypy from running on the tests in the COD AB PR, there were some super annoying issues with pytest fixtures and I thought it's probably not needed on the tests anyway. I should have flagged it more explicitly though, sorry guys. If you think we should put it back I will try to sort ou the fixture issues!

PR #29, opened 2021-11-09
ignore dsstore

Tinkaa:
Ignore Mac specific DS_Store files

Thought this is not worth mentioning in the changelog but correct me if I am wrong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ dmypy.json
 
 # PyCharm
 .idea/
+
+#Mac DSStore
+.DS_Store


caldwellst:
I would say don't put in change log, but on side note, here's how I configured my Git on MacOS to ignore all unnecessary files, described [here](https://sourabhbajaj.com/mac-setup/Git/gitignore.html) so I don't have to do it in each project. Uses GitHub's MacOS recommended `.gitignore`.

```shell
> curl https://raw.githubusercontent.com/github/gitignore/master/Global/macOS.gitignore -o ~/.gitignore_global
> git config --global core.excludesfile ~/.gitignore_global
```

I called it `.gitignore_global` instead of just `.gitignore` as makes it easier to quickly locate with Alfred.

turnerm (approved):

Issue #4, opened 2021-10-14
Automate doc building


turnerm:
Docs are hosted on github pages from the `gh-pages` branch, this deployment can be automated with github actions, see [here](https://betterprogramming.pub/auto-documenting-a-python-project-using-sphinx-8878f9ddc6e9) and [here](https://tech.michaelaltfield.net/2020/07/18/sphinx-rtd-github-pages-1/)

mcarans:
Following from our Slack chat, I looked at multiple doc tools and for the HDX Python libraries settled on pydoc-markdown. All the configuration for pydoc-markdown goes into pyproject.toml. It supports things like readthedocs etc., but I like its simplest option, a single markdown file in the root of the repo, which saves having to use GitHub pages eg.

```
[[tool.pydoc-markdown.loaders]]
type = "python"
search_path = ["src"]
packages = ["aatools"]

[tool.pydoc-markdown.renderer]
type = "markdown"
filename = "apidocs.md"

[tool.pydoc-markdown.renderer.source_linker]
type = "github"
repo = "OCHA-DAP/pa-aa-toolbox"

```

and 

```
legacy_tox_ini = """
...
[testenv:docs]
wheel_build_env = py38
deps =
    pydoc-markdown
commands =
    pydoc-markdown
...
"""
```

mcarans:
I changed my mind and investigated readthedocs :-) 

The toml looks like this
```
[[tool.pydoc-markdown.loaders]]
type = "python"
search_path = ["src"]
packages = ["hdx.utilities"]

[tool.pydoc-markdown.renderer]
type = "mkdocs"
output_directory = "docs/build"

[tool.pydoc-markdown.renderer.mkdocs_config]
site_name = "HDX Python Utilities"
theme = "readthedocs"
repo_url = "https://github.com/OCHA-DAP/hdx-python-utilities"

[[tool.pydoc-markdown.renderer.pages]]
title = "Home"
name = "index"
source ="docs/main.md"

[[tool.pydoc-markdown.renderer.pages]]
title = "API Documentation"
contents = [ "*" ]

[tool.pydoc-markdown.renderer.markdown.source_linker]
type = "github"
repo = "OCHA-DAP/hdx-python-utilities"
```

You also need a .readthedocs.yml (unfortunately not in toml) with:
```
version: 2
mkdocs: {}  # tell readthedocs to use mkdocs
python:
  version: 3.7
  install:
  - method: pip
    path: .
[...]

PR #35, opened 2021-11-17
Remove doc generation hook

### LINKED ISSUE #34: README documentation needs updating
The README documentation on the develop page needs updating. It currently references `python setup.py ...` in multiple instances even though the `setup.py` file is no longer being used.

On a similar note, running `pre-commit run --all-files` produces empty documentation for certain modules, such as everything in `utils`.

Manually specifying the rebuild using the below code works in generating the documentation for those modules:

```shell
sphinx-build -b html -d docs/build/doctrees docs/source docs/build/html
```

I am not very familiar with the setup we are using now and what's ideal, so raising this issue so others can input.

turnerm:
Closes #34. 

The pre-commit hook to generate the docs was not really working -- the actual docstrings were not showing up. I couldn't really figure out why; as far as I can tell the hook is running the following command:
```
sphinx-build -b html -d docs/build/doctrees docs/source docs/build/html
```
which does work on its own.

At any rate it's not crucial to have as a hook, since none of the changes it makes are actually committed; it was more for user convenience. For now, the user can just run the command by hand. I've added back the docs requirements file to permit this. 

Auto building will be addressed in #4.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,15 +40,6 @@ repos:
     rev: v0.29.0
     hooks:
       - id: markdownlint
-  - repo: https://github.com/thclark/pre-commit-sphinx
-    rev: 0.0.1
-    hooks:
-      - id: build-docs
-        additional_dependencies:
-          - sphinxcontrib-apidoc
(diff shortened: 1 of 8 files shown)


caldwellst (approved):
All looks great Monica, thanks for this!

PR #36, opened 2021-11-22
ECMWF API seasonal processing

turnerm:
A first crack at the processing of the ECMWF API seasonal data. This data differs from the CDS data in the following ways:
- there is no step coordinate, only a time coordinate
- the time coordinate has length 7

I assume that the time coordinate is just leadtime 0 - 7 months. So I created a step coordinate, and converted the time coordinate to be only the 0th entry. 

To be fully consistent with CDS, I think the 0th step would need to be removed, let me know if you think that should be done.

(The diff is currently very large because I've merged the latest `develop` into this branch)


Tinkaa:
Okay sorry I have to fix the merging mess I think I created by merging develop also in `ecmwf-seas-download`. 

Regarding the leadtimes: don't remove leadtime 0. This corresponds to leadtime 1 in the CDS data. However, it is correct that the API has one more leadtime. I am still confused by this as well. I think the data provided on the API just predicts one more month ahead in the future, but I am not 100% sure.. If that is the case I would just keep all the data and thus the data from the API would have one more leadtime than that from CDS. 

The other thing is on numbering the leadtimes. With CDS a leadtime of 1 refers to the month it is published in. So this is what the notebooks now assume. However, I do agree with your approach that it is a lot more logical to basically substract 1 from the leadtimes such that a leadtime of 0 refers to the month the forecast is published in. 
I am doubting a bit now as when we do that now it might cause a bit more struggle with the notebooks. What would you think of either 1) stick to the range of leadtimes of 1 to 7 for now but make an issue to change that later or 2) use 0 to 6 in his code but add a little conversion in `pa-aa-anticipatory-action` to change the range from 0 to 6 to 1 to 7 when loading the file?

Tinkaa:
Except for that no code comments

turnerm:
I'm trying to figure out the issue with the lead time. So looking at e.g. [this paper](https://drive.google.com/file/d/1q06A7lCLzNBIcMfJA0NDBctRXqe_MkjI/view?usp=sharing) indeed it looks like SEAS5 has up to 7 months lead time. But then on [CDS](https://cds.climate.copernicus.eu/cdsapp#!/dataset/seasonal-monthly-single-levels?tab=form) only up to 6 months is available. My guess is that what's available on CDS is slightly different, i.e. since it's technically [C3S](https://confluence.ecmwf.int/display/CKB/Seasonal+forecasts+and+the+Copernicus+Climate+Change+Service) that's why there are only 6 months lead time (and also why the resolution is lower). 

I'm a bit confused by [this table](https://confluence.ecmwf.int/display/CKB/Detailed+list+of+parameters). Does selecting ECMWF and system 5 in the API get you SEAS5? 

> The other thing is on numbering the leadtimes. With CDS a leadtime of 1 refers to the month it is published in. So this is what the notebooks now assume. However, I do agree with your approach that it is a lot more logical to basically substract 1 from the leadtimes such that a leadtime of 0 refers to the month the forecast is published in. I am doubting a bit now as when we do that now it might cause a bit more struggle with the notebooks. What would you think of either 1) stick to the range of leadtimes of 1 to 7 for now but make an issue to change that later or 2) use 0 to 6 in his code but add a little conversion in pa-aa-anticipatory-action to change the range from 0 to 6 to 1 to 7 when loading the file?

The reason I had indexed with 0 is because I didn't realize SEAS5 actually had 7 months, so I assumed the first entry was actually the model. But in fact I think your approach is much clearer, and I think we could just use 1 to 7 without making an issue to change it in the future. At any rate I will change it for now!

Last thing: it might be good to to a quick sanity check between the ECMWF API and CDS data, i.e. aggregating the ECMWF API data and checking that it is relatively consistent. Not sure how they aggregate but the mean should perhaps work.

Tinkaa:
>My guess is that what's available on CDS is slightly different, i.e. since it's technically C3S that's why there are only 6 months lead time (and also why the resolution is lower).

Would guess so as well. Just made a ticket to CDS to double-check with them. 

>I'm a bit confused by this table. Does selecting ECMWF and system 5 in the API get you SEAS5?

You mean in the CDS API? Then yes I am 99% sure about that. Is there anything that hints to you that that might not be the case? 

> I think we could just use 1 to 7 without making an issue to change it in the future. At any rate I will change it for now!

Perfect

>Last thing: it might be good to to a quick sanity check between the ECMWF API and CDS data, i.e. aggregating the ECMWF API data and checking that it is relatively consistent. Not sure how they aggregate but the mean should perhaps work.

Yes we should. I think the cleanest method would be to in the download code of the ECMWF API in `server.execute()` set `grid` to `1/1` instead of `0.4/0.4`. I remember I experimented a little with it and that it didn't fully overlap but I think I deleted all of that both physically and in my memory. So if you like to look into that a little that would be amazing, else I will do it later

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
```diff
@@ -138,3 +137,62 @@ def download_date(
...
+    #  them explicitly by looping over dates
+    filepath_list = [
+        filename
+        for filename in _get_output_path(raw_dir).iterdir()
+        if filename.is_file()
+    ]
+    output_filepath = (
+        _get_output_path(processed_dir) / f"{iso3}_{SEAS_DIR}_{PRATE_DIR}.nc"
+    )
+
```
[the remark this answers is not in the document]
turnerm: Done!

Tinkaa:
@turnerm added functionality to download for a list of dates, and to check file existence. Could you do a review?

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
```diff
@@ -42,14 +45,16 @@
...
 
-def download_date(
+def download(
     # question: do we want to assume an iso3 here?
     # might want to e.g. download for a multi-country region or only a city
     iso3: str,
-    date_forec: Union[str, date],
     ecmwf_dir: Union[str, Path],
     iso3_gdf: gpd.GeoDataFrame,
+    date_list: List[Union[str, date]] = None,
```
turnerm, line 54: I feel like a list of dates is a fairly complex parameter for the user to provide. For the GloFAS API for instance, I just had parameters for the start and end dates, which can be set to the extent of the dataset

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
```diff
@@ -42,14 +45,16 @@
...
     area: Area = None,
+    clobber: bool = False,
+    grid: float = 0.4,
```
turnerm, line 57: I should have added this in when I did it, but if we want to allow the user to vary the grid, it should maybe be included in the filename

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
```diff
@@ -61,45 +66,74 @@ def download_date(
...
+        date_list = [
+            d.strftime("%Y-%m-%d")
+            for d in pd.date_range(
+                start="1992-01-01", end="2021-04-01", freq="MS"
+            )
+        ]
+    for date_forec in date_list:
+        if not isinstance(date_forec, date):
+            date_forec = date.fromisoformat(date_forec)
+        # TODO: it would probably be safer to also include the boundary coords
```
turnerm, line 108: hm actually that's not a bad idea. If the filenames are all "under the hood" then the length doesn't matter too much. It's just tricky if the coords are long floats.

turnerm (commented):
Looks good! My main comment is about the date list parameter. Do you think providing a max and min date could work instead?

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
Tinkaa: Maybe we could round them to two decimals?

Tinkaa: And then also round the coords to two decimals before making the API request?

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
Tinkaa: Fair point. Changed it

src/aatoolbox/datasources/ecmwf_api/ecmwf_api_download_seasonal_monthly.py
```diff
@@ -109,11 +112,16 @@ def download(
         # TODO: it would probably be safer to also include the boundary coords
         # in the filename.. Just that the filename then gets massive
         output_filename = (
-            f"{iso3}_{SEAS_DIR}_{PRATE_DIR}_{date_forec.strftime('%Y-%m')}.nc"
+            f"{iso3}_{SEAS_DIR}_{PRATE_DIR}_{date_forec.strftime('%Y-%m')}"
         )
+        if area is not None:
+            output_filename += "_{area.get_filename_repr()}"
```
Tinkaa, line 118: @caldwellst just tagging you here because I  know you were adding coordinates to the ARC filenames, so that you know that that is now generalized in the toolbox so should be easy to use when refactoring ARC to toolbox

Tinkaa:
@turnerm Looks good to me! Regarding the grid resolution, makes sense to me to make it a global constant and then doesn't have to be included in the filename anymore I would say. Will try to find someone with ecmwf api experience to do a review of the code as well.

Tinkaa:
Close in favor of #37

PR #39, opened 2021-12-09
refactor area to utils

Tinkaa:
Refactor the area functionality from `cds` to `utils` and add functionality to create a string for filenames.

This is used by the ecmwf api code.


src/aatoolbox/datasources/cds/area.py
```diff
@@ -1,8 +1,8 @@
-"""Geographic area for the CDS API.
+"""Create geographic area for input to the CDS API.
```
turnerm, line 1: You can take out CDS since this is now becoming more general-purpose

src/aatoolbox/datasources/cds/area.py
```diff
@@ -12,10 +12,9 @@
...
 """
 from collections import namedtuple
-from typing import Dict, List, Union
+from typing import Dict, List
 
-import geopandas as gpd
-import numpy as np
+from aatoolbox.utils.area import Area
 
 Station = namedtuple("Station", "lon lat")
```
turnerm, line 19: This can be removed

src/aatoolbox/datasources/cds/area.py
```diff
@@ -32,83 +31,31 @@
...
-        return (
-            function(coord / round_val) * round_val
-            + offset_factor * offset_val
-        )
+    if round_coords:
+        area.round_area_coords(offset_val=offset_val)
+    return [area.north, area.west, area.south, area.east]
 
 
 class AreaFromStations(Area):
```
turnerm, line 61: Since there is only one subclass, I would just make this the `__init__` of `Area` and only have one class. Also maybe we should think about changing the name to make it more descriptive? `Area` is so general. Maybe `GeographicArea`?

src/aatoolbox/utils/area.py
```diff
@@ -0,0 +1,121 @@
+"""
```
turnerm, line 1: Ah wait, now I'm confused, why are there two area files?

turnerm (commented):
Looks good, but now I'm confused about there being two files. Also, make sure you update the tests to reflect these changes.

src/aatoolbox/datasources/cds/area.py
Tinkaa: As discussed, only using the `utils/area.py` class now
I would agree `Area` is a bit general. I am now also thinking that it is more the bounding box than the area. Would `GeographicBoundingBox` make sense? Or too long?

Tinkaa:
@turnerm updated name and tests and removed `cds`. Let me know what you think

src/aatoolbox/utils/geoboundingbox.py
```diff
@@ -0,0 +1,122 @@
...
+            else:
+                return f"p{coord:.{p}f}"
+
+        return (
+            f"N{_str_format(self.north)}S{_str_format(self.south)}"
+            f"E{_str_format(self.east)}W{_str_format(self.west)}"
+        )
+
+
+class GeographicBoundingBoxFromShape(GeographicBoundingBox):
```
turnerm, line 100: I would remove this subclass and just add this functionality to the parent class `__init__`, since we don't have the `AreaFromStations` functionality anymore, there's no point in having a single child class

turnerm (approved):
One last change, just to make it a single class!

src/aatoolbox/utils/geoboundingbox.py
Tinkaa: As discussed changed the class to a `@classmethod` within `GeographicBoundingBox`. Does this look correct to you?

turnerm: Yes looks great, merge away!

Issue #40, opened 2021-12-14
Seasonal forecasts: time and leadtime conventions


Tinkaa:
We use seasonal forecasts from different providers. These providers use different conventions in their `time` and `leadtime` parameters. This leads to confusion. For example in IRI's forecast a leadtime of 1 refers the projected period being the month of the `time` parameter + 1 month. While with ECMWF a leadtime of 1 refers to the projected beriod being the month of the `time` parameter. `time` generally refers to the date the forecast is published. 

We should be clear on the conventions we use and have them documented somewhere. 

Issue based on disccusion [here](https://github.com/OCHA-DAP/pa-anticipatory-action/pull/248)

Tinkaa:
Also rethink the naming of the parameters. `time` maybe could be rebranded to `pub_time`. And we use `step` at some points which could be `leadtime`

PR #43, opened 2021-12-16
ecmwf: add start time coord

Tinkaa:
Based on our discussion [here](https://github.com/OCHA-DAP/pa-anticipatory-action/pull/248) and #40, I attempted to add an additonal coord `start_forecast_time` which indicates the start date at which the forecast is valid and is dependent on `time` and `step`. It almost works but due to only having one date for the realtime dataset it doesn't merge nicely. 

I got frustrated and couldn't find a solution, so wondering if you know one? And if it is worth making it work when one of the datasets has only one entry, or that we should just assume there are several..

### DIFF
--- a/src/aatoolbox/datasources/ecmwf/realtime_seas5_monthly.py
+++ b/src/aatoolbox/datasources/ecmwf/realtime_seas5_monthly.py
@@ -151,7 +151,9 @@ class EcmwfRealtime(DataSource):
             preprocess=lambda d: self._preprocess(d),
             # time refers to the publication month
             # forecastMonth to the leadtime in months, which is indexed 1-7
-            backend_kwargs={"time_dims": ("time", "forecastMonth")},
+            backend_kwargs={
+                "time_dims": ("time", "forecastMonth", "verifying_time")
+            },
         ) as ds:
             ds.to_netcdf(output_filepath)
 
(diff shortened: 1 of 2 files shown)


turnerm:
I'm not totally sure I understand exactly what you are trying to do, could you write explicitly which coordinate you're trying to create and what they would look like (and what the final dataset would look like)? Note that I would try to make the two ECMWF sources as similar as possible by the end of this process, if that makes sense.

src/aatoolbox/datasources/ecmwf/realtime_seas5_monthly.py
```diff
@@ -173,7 +175,18 @@ def get_processed_path(self):
...
+        ds_date = ds_date.rename(
+            {
+                "forecastMonth": "step",
+                # arghh because we only have one time step
+                # till now, this doesn't work nicely since
+                # it is then only dependent on step instead
+                # of time and step. This then gives errors
+                # when trying to merge with the api data
+                # TODO: help I don't know how to solve this..
+                # "verifying_time":"start_forecast_time"
```
turnerm, line 187: If there is a merging issue you may need to do add this coordinate to `expand_dims`

Tinkaa:
Closing for now, as #37 should be finished first, and that code has been refactored substantially in the mean time.

PR #44, opened 2021-12-22
Publish on PyPI

turnerm:
Publish `aa-toolbox` on PyPI. The changes include:
- GitHub Action to publish on PyPI every time a tag is pushed (so on a version release), and on TestPyPI on every commit. Note that the PyPI publication bit has not yet been tested, I'm not sure how to do that other than make a release and tag `main`. But I think the code should work because I copied the [guide from PyPI](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)
- Changed to using [`setuptools_scm`](https://github.com/pypa/setuptools_scm) for versioning, so that each commit would have a unique number for TestPyPI

You can see the all the releases on TestPyPI [here](https://test.pypi.org/project/aa-toolbox/#history). This way, if you are in dire need of the functionality in one of the external AA repos, you can always (temporarily) use TestPyPI and not need to wait for a release.


Tinkaa:
It looks great! Though tbh I am not an expert on this stuff so do you think we need to have e.g. Mike still have a look at it?

>Note that the PyPI publication bit has not yet been tested, I'm not sure how to do that other than make a release and tag main.

Cant we just do this? 


And do we need any documentation (possibly outside GitHub) on how (Test)PyPi works for the team? 
Not related to the code but one question I had: when is it publishted on `TestPyPi`? Cause I now see 4 versions but we definitely made more commits.

Tinkaa (approved):
It looks great! Though tbh I am not an expert on this stuff so do you think we need to have e.g. Mike still have a look at it?

> Note that the PyPI publication bit has not yet been tested, I'm not sure how to do that other than make a release and tag main.

Cant we just do this?

And do we need any documentation (possibly outside GitHub) on how (Test)PyPi works for the team?
Not related to the code but one question I had: when is it publishted on TestPyPi? Cause I now see 4 versions but we definitely made more commits.

turnerm:
@Tinkaa 

> It looks great! Though tbh I am not an expert on this stuff so do you think we need to have e.g. Mike still have a look at it?

Good idea, I will add him!

>> Note that the PyPI publication bit has not yet been tested, I'm not sure how to do that other than make a release and tag main.

> Cant we just do this?

Yes I was thinking once this is merged we can just do a release. 

> And do we need any documentation (possibly outside GitHub) on how (Test)PyPi works for the team?

That's a good point, I can just add some to the README for now.

> Not related to the code but one question I had: when is it publishted on TestPyPi? Cause I now see 4 versions but we definitely made more commits.

Ah so only the commits from now on will be published on TestPyPI, it won't do it retroactively. And some of those versions are tests so the numbers will start looking more linear form now on I think.

Tinkaa:
Perfect, sounds great!

mcarans (approved):
This is an innovative solution to the versioning problem that solves the problem of getting the Git tag and published version on PyPI lined up without adding extra steps. 

The only issues I ran into were:
The fiona dependency gave the error: "A GDAL API version must be specified. Provide a path to gdal-config using a GDAL_CONFIG environment variable or use a GDAL_VERSION environment variable"

The README could have a line that says to install GDAL. In my case (Linux Mint), I followed some online instructions:
```
sudo add-apt-repository ppa:ubuntugis/ppa && sudo apt-get update
sudo apt install libgdal-dev
```

I then ran into this [issue in rasterio](https://github.com/rasterio/rasterio/issues/2333) which affects Python 3.10. Until the fix is released, then it is necessary to either build the latest Cython (I think) or use Python <= 3.9. I chose the latter and found Python 3.8 works fine.

Other than these minor issues, looks great!

turnerm:
@mcarans thanks for the helpful comments, I've made a note about GDAL in the README and (temporarily) limited the Python version to <3.10

PR #45, opened 2022-01-03
Automate testing

turnerm:
Automate unit test running in github actions. 

@mcarans I implemented a "matrix" to use different Python versions, is there a reason why you don't do that? Does it mess up the reports in your experience? 

Removed Python 3.6 for now while I try to figure out an error with the pacakge installation.

### DIFF
--- a/.github/workflows/run-python-tests.yaml
+++ b/.github/workflows/run-python-tests.yaml
@@ -4,12 +4,15 @@ jobs:
   build:
     name: Run unit tests
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.6', '3.7', '3.8', '3.9']
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: ${{ matrix.python-version }}
     - name: Install GDAL
       run: |
         sudo apt-get update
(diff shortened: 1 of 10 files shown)


mcarans (approved):
This looks good to me.  If you'd like to have less files, the contents of tox.ini can be put into pyproject.toml where I've put XXXX:
```toml
[tool.tox]
legacy_tox_ini = """
XXXX
"""
```

You could consider using tox-wheel (which now has my PR merged and a new version released) to test using the built binary - in which case you can set it up as in the tox section here: https://github.com/OCHA-DAP/hdx-python-api/blob/main/pyproject.toml

The error with test with Python 3.6 should be fixed with version 5.4.9 of HDX Python API.

turnerm:
@mcarans thanks so much for the great suggestions. I've switched to `wheel`, and updated HDX Python API to the latest version. 

The only thing that I'm not totally happy with, although it's probably not worth spending more time on, is that I would prefer the Codecov upload to be in the "publish" job section. For example in [this](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/45/commits/55a7c37a3c673b968c5513007701e7b56592121c) commit, the [action itself](https://github.com/OCHA-DAP/pa-aa-toolbox/runs/4705968022?check_suite_focus=true) succeeded, but Codecov [struggled](https://codecov.io/github/OCHA-DAP/pa-aa-toolbox/commit/55a7c37a3c673b968c5513007701e7b56592121c) to process the report.

PR #46, opened 2022-01-04
bugfix: duplicate test PyPI versions

turnerm:
`setuptools_scm` uses the distance from the tag to auto generate the version number. As I feared, the distance does not take into account parallel branches, so you do get some overlap in version numbers if you version every commit. Instead, I propose to only publish to Test PyPI for any merges to `main` or `develop`.


mcarans (approved):
This looks good to me. I think there is unlikely to be a need to publish to test pypi very frequently.

PR #58, opened 2022-02-22
Release/0.4.0

Tinkaa:
I want to release the new code which includes IRI functionality. 

What I understand from the readme, all I have to do is update the changelog and then it will be directly published to PyPi. Is that correct? 

Also, should hereafter main be merged to develop or smth?


Tinkaa:
Closing as we decided to not do a new release on PyPI yet

PR #59, opened 2022-02-24
update testpypi install command

Tinkaa:


turnerm (approved):
Thanks for taking care of this!

PR #75, opened 2022-03-01
add cod config

Tinkaa:
Add config for COD (=DRC). 

Should this be added to the changelog as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to
 
 - Implemented unit testing for `AatRaster` module with full coverage.
 - Implemented downloading and processing for IRI seasonal precipication forecast.
-- Added config for COD
+- Added config for DRC (of which the iso3 is COD)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 
 - Implemented unit testing for `AatRaster` module with full coverage.
 - Implemented downloading and processing for IRI seasonal precipication forecast.
+- Added config for COD
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


src/aatoolbox/config/countries/cod.yaml
```diff
@@ -0,0 +1,7 @@
+iso3: cod
+codab:
+  hdx_address: cod-ab-cod
+  hdx_dataset_name: cod_admbnda_rgc_itos_20190911_SHP.zip
+  #the adm2 layer name is cod_admbnda_adm2_rgc_20190911.shp
```
turnerm, line 5: Does this mean that admin 2 doesn't follow the same structure as 0-1? I was afraid this might happen at some point. The way we could handle this for now is by adding admin 2 as a custom layer name, see the Nepal config as an example. Then add a comment in the config file.

Tinkaa: Oeh awesome! Like I did now?

src/aatoolbox/config/countries/cod.yaml
```diff
@@ -0,0 +1,9 @@
+iso3: cod
+codab:
+  hdx_address: cod-ab-cod
+  hdx_dataset_name: cod_admbnda_rgc_itos_20190911_SHP.zip
+  layer_base_name: cod_admbnda_adm{admin_level}_rgc_itos_20190911.shp
+  admin_level_max: 2
```
turnerm, line 6: This should be 1 since 2 is in the custom layers

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to
 
 - Implemented unit testing for `AatRaster` module with full coverage.
 - Implemented downloading and processing for IRI seasonal precipication forecast.
+- Added config for COD
```
turnerm, line 18: Maybe mention that you mean DRC just to differentiate from the COD ABs

turnerm (approved):

Issue #62, opened 2022-02-27
check_file_existence decorator


turnerm:
The `check_file_existence` decorator should be fixed to handle instance methods

PR #79, opened 2022-03-15
Add wrapt decorator to check file existence for instance methods

caldwellst:
Attempt to address #62. Uses [wrapt](https://wrapt.readthedocs.io/en/latest/quick-start.html) which allows one standard decorator for instance methods, class methods, functions, etc. by ensuring bound object not passed to args.

However, I couldn't eventually wrap my head around how to implement your typing cast in the same way as in the original. As well, would need to implement some tests on the decorator during changing. Putting here because is WIP, but taking me a while to wrap my head around as I'm quite rusty now.

### DIFF
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -553,6 +553,8 @@ wheel==0.37.0
     #   -r requirements/requirements.txt
     #   libhxl
     #   pip-tools
+wrapt==1.14.0
+    # via -r requirements/requirements.txt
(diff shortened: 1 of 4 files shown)


caldwellst:
I had something working and clearly it's not now. Just not really understanding it at the moment, so no need to review yet I think, although an explanation would be greatly appreciated if you get time ha!

turnerm:
To revisit at a later time

Issue #61, opened 2022-02-27
DataSource should be an abstract base class 


turnerm:
`DataSource` should be made an [abstract base class](https://docs.python.org/3/library/abc.html) as it is an empty class that is not meant to be instantiated by users.

If the signatures of `download`, `process`, and `load` are consistent then they can also be defined in the base class. This would also be a good place to document the methods generally and the order in which they should be run, etc.

PR #80, opened 2022-03-15
Datasource abstract

caldwellst:
Addresses #61. Creates the DataSource as an abstract base class with abstract methods. Some issues that arise are:

- All abstract methods must be overridden for a derived class to be instantiated. Thus, `CodAB` requires a different abstract base class since it lacks a `process()` method.
- There are varying parameters that will be passed to the methods `download()`, `process()`, and `load()` in derived classes (like in `CodAB`, where `load()` takes the admin level as an argument). This is an issue with [PEP8](https://stackoverflow.com/questions/42778784/abstract-classes-with-varying-amounts-of-parameters) caught by mypy and has to be excepted. The only alternative would be not to have abstract methods for the above, and just have an abstract method for `__init__`, which would be cleaner for having a single `DataSource` abstract class, but lose the ability to force derived classes to have the other 3 methods explicitly.

### DIFF
--- a/src/aatoolbox/datasources/codab/codab.py
+++ b/src/aatoolbox/datasources/codab/codab.py
@@ -28,8 +28,6 @@ from aatoolbox.utils.io import check_file_existence
 
 logger = logging.getLogger(__name__)
 
-_MODULE_BASENAME = "cod_ab"
-
 
 class CodAB(DataSource):
     """
@@ -43,11 +41,11 @@ class CodAB(DataSource):
 
     def __init__(self, country_config: CountryConfig):
         super().__init__(
-            country_config, module_base_dir=_MODULE_BASENAME, is_public=True
+            country_config, module_base_dir="cod_ab", is_public=True
         )
         self._raw_filepath = (
             self._raw_base_dir
-            / f"{self._country_config.iso3}_{_MODULE_BASENAME}.shp.zip"
+            / f"{self._country_config.iso3}_{self._module_base_dir}.shp.zip"
         )
 
     def download(self, clobber: bool = False) -> Path:
@@ -86,9 +84,6 @@ class CodAB(DataSource):
         """
         logger.info("`process()` method not implemented for CodAB.")
 
-    # next line ignored to avoid PEP8 error
-    # described here:
-    # https://stackoverflow.com/questions/42778784/abstract-classes-with-varying-amounts-of-parameters
     def load(self, admin_level: int) -> gpd.GeoDataFrame:  # type: ignore
         """
         Get the COD AB data by admin level.
--- a/src/aatoolbox/datasources/codab/codab.py
+++ b/src/aatoolbox/datasources/codab/codab.py
@@ -21,14 +21,14 @@ import geopandas as gpd
 from fiona.errors import DriverError
 
 from aatoolbox.config.countryconfig import CountryConfig
-from aatoolbox.datasources.datasource import DataSource
+from aatoolbox.datasources.datasource import DataSourceNoProcess
 from aatoolbox.utils.hdx_api import load_dataset_from_hdx
 from aatoolbox.utils.io import check_file_existence
 
 _MODULE_BASENAME = "cod_ab"
 
 
-class CodAB(DataSource):
+class CodAB(DataSourceNoProcess):
     """
     Work with COD AB (administrative boundaries).
 
@@ -75,7 +75,7 @@ class CodAB(DataSource):
             clobber=clobber,
         )
 
-    def load(self, admin_level: int) -> gpd.GeoDataFrame:
+    def load(self, admin_level: int) -> gpd.GeoDataFrame:  # type: ignore
         """
         Get the COD AB data by admin level.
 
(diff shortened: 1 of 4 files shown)


turnerm:
Thanks so much for tackling this Seth, it looks great so far. Some things that I should have specified in the issue itself (so sorry that I didn't):

- It would be great if it could use the Python [ABC library](https://docs.python.org/3/library/abc.html)
- I was thinking that `_MODULE_BASENAME` could be a required property (since it's needed to set the filepath)
- For classes with no `process` method, I'm wondering if it's just easier to have a single base class, and then for `process` it can just log a message saying "not implemented yet" or something like that
- The idea would be also to hopefully standardize constructor arguments. In general most classes would need a country config and geographical area. We'd have to think about how to deal with custom arguments and / or classes that don't need either of those two.

src/aatoolbox/datasources/codab/codab.py
```diff
@@ -75,7 +78,18 @@ def download(self, clobber: bool = False) -> Path:
...
         )
 
-    def load(self, admin_level: int) -> gpd.GeoDataFrame:
+    def process(self):
+        """
+        Process COD AB data.
+
+        Method not implemented.
+        """
+        logger.message("`process()` method not implemented for CodAB.")
```
turnerm, line 87: Should be `logger.info`

turnerm (commented):
> This is using the ABC library!

:facepalm:  so sorry, I am blind. Can I blame an empty stomach?

> I think this is already captured, right? The DataSource class requires module_base_dir when instantiated, which is what _MODULE_BASEDIR is passed to. Given _MODULE_BASEDIR is just a global variable set in the modules themselves, not sure there's a cleaner way to check for its presence than that it's passed into the instantiation?

I was thinking that you could do something like:
```python
@property
@abstractmethod
def _module_basename(self):
    pass
```
also for `_is_public`. Then it forces the creation of these properties every time you make a `DataSource` sub class. Because the problem with the `_MODULE_BASEDIR` constant is that we're not enforcing that it be created for each class, if you know what I mean. 

> Similarly, country_config is also required in the constructor. This gets to the issue of how wide we want the base class to be. If we expand to include constructors, then for classes that don't require or use those arguments, like geographical area, I think they'd have to be optional. But, the geographical area for each API is often different, such as points for GLOFAS, bounding box for IRI, different coordinates for ARC2, and so on. You could in that specific data sources class call them by their own name, such as geo_bounding_box, and then pass back to the abstract class as geographic_area? I just don't think is necessarily worth it.

I agree with you, although I will just note that GloFAS and IRI do use the same geo bounding box as input, which is then manipulated differently within the classes themselves. (not sure about ARC2).

turnerm (approved):
Oops sorry had meant to select approve

src/aatoolbox/datasources/datasource.py
```diff
@@ -53,3 +59,18 @@ def _get_base_dir(self, is_public: bool, is_raw: bool) -> Path:
             / self._country_config.iso3
             / self._module_base_dir
         )
+
+    @abstractmethod
+    def download(self):
```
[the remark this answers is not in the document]
caldwellst: Have made `clobber` a required parameter. However, the `*args` and `**kwargs` don't help signature compatibility sadly!

src/aatoolbox/datasources/codab/codab.py
```diff
@@ -75,7 +76,15 @@ def download(self, clobber: bool = False) -> Path:
...
         )
 
-    def load(self, admin_level: int) -> gpd.GeoDataFrame:
+    def process(self, *args, **kwargs):
+        """
+        Process COD AB data.
+
+        Method not implemented.
+        """
+        logger.info("`process()` method not implemented for CodAB.")
```
AlessioTonioni, line 85: nit: if it's not required I would lean towards modifying the log info to explicitly say it )rather than not implemented which seems to imply that it will be implemented at some point)

AlessioTonioni (commented):
Hello, so sorry to check this only now. Looks great!
Some comments on the points raised in the discussion:

* `process()`: the way it is implemented right now I think it's completely fine. An alternative could have been to have `DataSource` as the base class with only `download` and `load` and a class inheriting from it (DataSourceWithProcess), still abstract and adding the `process` method. This is probably an overkill in this context, but would allow concrete classes to inherit from one or the other depending if they need process or not. As I said the current solution I think it's completely fine.

* Different parameters for different concrete classes. One option is to have an ad hoc object containing the extra parameters specific for the subclass. For example `download` could take a single argument `parameters` containing any parameters needed to run with that specific subclass. The same can be implemented using [the `**kwargs` argument](https://book.pythontips.com/en/latest/args_and_kwargs.html). I'm not a particular fan of either but they are available options.

turnerm:
@AlessioTonioni thanks so much for your comment, very helpful and we'll keep it in mind as we keep developing!

src/aatoolbox/datasources/codab/codab.py
turnerm: Yes agreed! (but in this case we would like to implement a processing step in the future)

PR #81, opened 2022-03-17
add is_global

Tinkaa:
Adding an `is_global` par to the DataSource class to enable loading from global datasources.

Not sure what to do when the raw datasource is in global and we process it to only contain country data (or maybe we shouldn't do that anyway?)

### DIFF
--- a/src/aatoolbox/datasources/datasource.py
+++ b/src/aatoolbox/datasources/datasource.py
@@ -4,6 +4,8 @@ from pathlib import Path
 from aatoolbox.config.countryconfig import CountryConfig
 from aatoolbox.config.pathconfig import PathConfig
 
+_GLOBAL_DIR = "glb"
+
 
 class DataSource:
     """
@@ -25,19 +27,27 @@ class DataSource:
         country_config: CountryConfig,
         module_base_dir: str,
         is_public: bool = False,
+        is_global: bool = False,
     ):
 
         self._country_config = country_config
         self._module_base_dir = module_base_dir
         self._path_config = PathConfig()
         self._raw_base_dir = self._get_base_dir(
-            is_public=is_public, is_raw=True
+            is_public=is_public,
+            is_global=is_global,
+            is_raw=True,
         )
         self._processed_base_dir = self._get_base_dir(
-            is_public=is_public, is_raw=False
+            is_public=is_public, is_global=is_global, is_raw=False
         )
 
-    def _get_base_dir(self, is_public: bool, is_raw: bool) -> Path:
+    def _get_base_dir(
+        self,
+        is_public: bool,
+        is_raw: bool,
+        is_global: bool,
+    ) -> Path:
         permission_dir = (
             self._path_config.public
             if is_public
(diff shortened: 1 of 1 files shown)


src/aatoolbox/datasources/datasource.py
```diff
@@ -25,19 +27,27 @@ def __init__(
...
+        is_global: bool = False,
     ):
 
         self._country_config = country_config
         self._module_base_dir = module_base_dir
         self._path_config = PathConfig()
         self._raw_base_dir = self._get_base_dir(
-            is_public=is_public, is_raw=True
+            is_public=is_public,
+            is_global=is_global,
```
turnerm, line 38: I wouldn't necessarily make this a instance parameter, but rather just a parameter that you can pass to `_get_base_dir` that can be used in the class itself (you'd have to just call `_get_base_dir` again). I'd imagine that you probably only want to use it for the raw data, then you'd want to keep the country iso3 for the processed data.

src/aatoolbox/datasources/datasource.py
```diff
@@ -25,19 +27,27 @@ def __init__(
...
-            is_public=is_public, is_raw=False
+            is_public=is_public, is_global=is_global, is_raw=False
         )
 
-    def _get_base_dir(self, is_public: bool, is_raw: bool) -> Path:
+    def _get_base_dir(
+        self,
+        is_public: bool,
+        is_raw: bool,
+        is_global: bool,
```
turnerm, line 49: Set this to `False` as default

turnerm (approved):
Looks good, just some minor changes

turnerm:
Adding @caldwellst as a reviewer to see what he thinks!

caldwellst:
Oooh was just thinking about this the other day, and this is a nice implementation. I am not sure, but think I'd agree about it not being an instance parameter, particularly because maybe there are cases where maybe even processed data would be processed at a global level.

As is, the argument needs adding to the documentation.

Long term, in terms of flexibility, should we make a "custom dir" argument? It might make sense to download data globally, regionally, or any other setup that isn't necessarily fitting into the ISO3 paradigm. And that could hold true for processing data as well, potentially. Maybe this a longer term issue, but yeah, feel that there's lots of cases where users and ourselves might want to process to a region rather than just a country, download for a wider area, or otherwise have the flexibility to use custom directories not tied into our current paradigm but still have the same capacity to access and process data.

caldwellst (approved):

turnerm:
> Long term, in terms of flexibility, should we make a "custom dir" argument? It might make sense to download data globally, regionally, or any other setup that isn't necessarily fitting into the ISO3 paradigm. And that could hold true for processing data as well, potentially. Maybe this a longer term issue, but yeah, feel that there's lots of cases where users and ourselves might want to process to a region rather than just a country, download for a wider area, or otherwise have the flexibility to use custom directories not tied into our current paradigm but still have the same capacity to access and process data.

Yeah that's a good point. I guess my only worry is that the this directory could become a bit chaotic, unless we settle on very strict definitions on regional names to use and what constitutes a certain region. In that sense I prefer putting anything not specific to a country in `glb`. I would say that once we encounter a clear example where this system is not working well, then we can address it.

Tinkaa:
Nice discussion. I get now why it shouldn't be an instance par and agree. 

Good point Seth, I am really on the edge about this one. I think in general it is good-enough to put anything non-country into iso3, but external users might not like that. I think if we would ever e.g. develop a regional framework, we could have a config with the regional name, instead of needing a custom string in this method. 

Do you have a concrete example in mind where we would need other divisions than iso3&glb?

caldwellst:
No clear example yet of what that could look like, and agree that maybe for now using `glb` is the best solution, and we can expand as necessary. I also chatted with Monica yesterday and am gonna work on just having an ability to flexibly get and set the filepaths, which might help resolve these issues, but can be part of a longer term effort.

PR #84, opened 2022-03-21
refactor fewsnet datasource

Tinkaa:
Refactor FewsNet downloading code to follow the subclass of DataSource structure. Big compliments to you how the toolbox now enables to write sooo much cleaner code! 

Two things that need to change, that is why I made it a draft, but apart from those it is ready for full review. 
1. I added fewsnet as an optional arg to the CountryConfig. However, then mypy starts complaining later on that `region_name` can be None. Do you know the best way to fix this? 
2. All tests are running, except `test_download_nodata` where I am getting the error 
`FAILED tests/datasources/test_fewsnet.py::test_download_nodata - ModuleNotFoundError: No module named ‘aatoolbox.datasources.fewsnet.fewsnet.Fewsnet’; ‘aatoolbox.datasources.fewsnet.fewsnet’ is not a package` 
I didnt manage to figure out why I get this error here and not for the other functions. Would you have an idea? (commented out this function for now).


src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,229 @@
...
-            # country - date is not available
-            logger.debug(
-                f"No zip data returned from url {url} "
-                f"check that the area and date exist."
+        Examples
+        --------
+        >>> download(date_pub="2020-10-01")
+        """
+        if (
+            self._country_config.fewsnet.region_name  # type: ignore
```
turnerm, line 115: I think I had this problem also with GloFAS. If I recall correctly, the way to fix it (which is actually the correct thing to do) is to check and warn / exit if `country_config.fewsnet` is `None` i.e. not implemented

turnerm (commented):
Hey Tinka, thanks so much for taking care of this. I haven't had a chance to go through it all but I wanted to quickly address your two points. (1) is addressed in the comments below, and for (2) I think you just need to add `__init__.py` ;)

src/aatoolbox/config/countryconfig.py
```diff
@@ -31,11 +31,45 @@ class CodABConfig(BaseModel):
...
+        Name of the region the country belongs to. Needed to download the
+        regional FEWS NET data
+    """
+
+    region_name: str
+
+    @validator("region_name")
+    def regionname_valid(cls, v):
+        """Check that regionname is one of the valid ones."""
+        valid_regionnames = [
```
Tinkaa, line 49: @turnerm hmm I am wondering how we can document clearly what the valid regionnames are.

Tinkaa:
🤦‍♀️ 🤦‍♀️  there was a typo in the nodata test that caused the error we were discussing above. So now everything is working, removing the draft status.

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
+        if not isinstance(date_pub, date):
+            date_pub = datetime.strptime(date_pub, _DATE_FORMAT)
+
+        # we prefer the country data as this more nicely structured
+        # thus first check if that is available
+        country_path = self._download_country(
+            date_pub=date_pub,
+            clobber=clobber,
+        )
+        if country_path is not None:
```
turnerm, line 116: I think it would be better if you raised an error in `_download_zip` and let it bubble up, instead passing `None` all the way back here. Then instead of all this logic here you can use try-catch.

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
+            country_data = self._download_zip(
+                url=url_country_date,
+                zip_filename=zip_filename,
+                output_dir=output_dir_country,
+            )
+        else:
+            country_data = output_dir_country
+        return country_data
+
+    def _download_region(
```
turnerm, line 162: There's a lot of overlap with `_download_country`, maybe they can be refactored a bit?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
+
+            try:
+                unzip(zip_file_path=zip_path, save_dir=output_dir)
+                logger.debug(f"Unzipped to {output_dir}")
+
+            except zipfile.BadZipFile:
+                # indicates that the url returned something that wasn't a
+                # zip, happens often and indicates data for the given
+                # country - date is not available
+                logger.debug(
```
turnerm, line 243: As mentioned above, I would re-raise this error and catch it in the case of `country` so that you can try the region.

tests/datasources/test_fewsnet.py
```diff
@@ -1,81 +1,96 @@
 """Tests for the FewsNet module."""
+from pathlib import Path
+
 import pytest
+from conftest import FAKE_AA_DATA_DIR, ISO2
+
+from aatoolbox.datasources.fewsnet.fewsnet import FewsNet
 
-from aatoolbox.datasources.fewsnet.fewsnet import download_fewsnet
+MODULE_BASENAME = "fewsnet"
```
turnerm, line 9: I started calling this `DATASOURCE_BASE_DIR` instead (just in the tests)

turnerm (commented):
Amazing job Tinka, the tests are excellent, you are a mocking pro! A couple of comments on the structure but overall I think it's looking really great.

One question I have is, is there going to be a `process` method? If not then we should think more about that discussion we had on Slack regarding how to structure the base class. Either way, I think you should merge in the latest `develop` because you'll need to create `process` method because of the abstract base class.

src/aatoolbox/config/countries/eth.yaml
```diff
@@ -3,3 +3,5 @@ codab:
   hdx_dataset_name: ETH Administrative Divisions Shapefiles.zip
   layer_base_name: eth_admbnda_adm{admin_level}_csa_bofed_20201008.shp
   admin_level_max: 2
+fewsnet:
+  region_name: east-africa
```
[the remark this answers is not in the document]
Tinkaa: Ohh yes I wanted to discuss. For some countries, e.g. Malawi, it is available but we haven't used it in the pilot. However, I would argue to still add it to the config. Would you agree?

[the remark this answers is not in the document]
turnerm: Definitely!

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
-    zip_filename_region = f"{output_dir_region.name}.zip"
-    if not output_dir_region.exists() or use_cache is False:
-        region_data = _download_zip(
-            url=url_region_date,
-            zip_filename=zip_filename_region,
-            output_dir=output_dir_region,
+        # FN data can be regional, and thus we save the raw data
+        # in "glb" instead of in the iso3 folder, even though part of it is at
+        # country level
+        self._raw_base_dir = self._get_base_dir(
```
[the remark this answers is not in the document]
Tinkaa: Yeah it might be a good idea (though not applicable here yet as we didn't implement `process`).
I am just wondering if we ever want to save processed global data, what you think?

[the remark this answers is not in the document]
Tinkaa: Anyway especially in view of other users it might be better to allow for that

src/aatoolbox/datasources/fewsnet/fewsnet.py
Tinkaa: I know I say this almost every PR, but thanks for making this code so much better!

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
+        clobber : bool, default = False
+            If True, overwrites existing raw files
+
+        Examples
+        --------
+        >>> download(date_pub="2020-10-01")
+        """
+        # convert to datetime if str
+        if not isinstance(date_pub, date):
+            date_pub = datetime.strptime(date_pub, _DATE_FORMAT)
```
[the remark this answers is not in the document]
Tinkaa: Yeah good point. I am rethinking and do you think it would make more sense to split `date_pub` into two input args `pub_year` and `pub_month`? As we never use the day of the month, so this might be confusing to the user?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,243 +32,218 @@
...
-        "west-africa",
-    ],
-    region_code: Literal["LAC", "EA", "CA", "SA", "WA"],
-    output_dir: Union[Path, str],
-    use_cache=True,
-) -> Path:
-    """
-    Retrieve the raw fewsnet data.
+# the URL uses a code, so map them
+_REGION_NAME_CODE_MAPPING = {
```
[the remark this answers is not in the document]
turnerm: Oh it's just bad to have to repeat fixed parameters like this, when there is e.g. big potential for a typo. A couple of ideas: 

- what if you put `_REGION_NAME_CODE_MAPPING` in `countryconfig.py` and imported it into `fewsnet.py`?
- what if you made `region_name_code_mapping` a parameter of the config class itself? Maybe I like this one!

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
turnerm: > I am just wondering if we ever want to save processed global data, what you think?

Hm yeah, if you can't think of any examples then I certainly can't. However, could be good to have for flexibility when subclassing DataSource in the pilot repos.

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
turnerm: Oo yes I like that idea a lot!

src/aatoolbox/config/countryconfig.py
```diff
@@ -31,11 +31,53 @@ class CodABConfig(BaseModel):
...
+    Parameters
+    ----------
+    region_name: str
+        Name of the region the country belongs to. Needed to download the
+        regional FEWS NET data
+    """
+
+    # values dictionary gets build in order attributes are listed
+    # so first define the dict before the region_name
+    region_name_code_mapping: Dict[str, str] = {
```
Tinkaa, line 46: @turnerm I now added the list of region name - code combinations here and then also assigned the region code in the config. Does this make sense to you?

Tinkaa:
Alrighty @turnerm thanks for all the feedback! Made quite some changes and few open questions. 

Only thing left is the documentation, which I will try to do Friday, but in case you have energy tomorrow to have a last look at the code, I am putting it up here.

Tinkaa:
Documentation is up as well!

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
-        "west-africa",
-    ],
-    region_code: Literal["LAC", "EA", "CA", "SA", "WA"],
-    output_dir: Union[Path, str],
-    use_cache=True,
-) -> Path:
-    """
-    Retrieve the raw fewsnet data.
+_VALID_PROJECTION_PERIODS_LITERAL = Literal["CS", "ML1", "ML2"]
+_VALID_PROJECTION_PERIODS = ["CS", "ML1", "ML2"]
```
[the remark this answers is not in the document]
caldwellst: What wasn't working for `__args__`? I got it to work fine just ignoring using `mypy`.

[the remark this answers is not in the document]
Tinkaa: yeah I first thought so as well, it is importing it fine. But then when it is actually using the `__args__` in the tests, it got an error for 3.6 that it couldnt find `__args__`.. Dont really understand why but didnt want to waste more time on it.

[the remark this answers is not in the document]
caldwellst: Oh very strange!

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
+            is_global_processed=False,
         )
-    if region_path is not None:
-        return region_path
-    else:
-        raise RuntimeError(f"No data found for {date_pub.strftime('%Y-%m')}")
 
+        self._iso2 = Country.get_iso2_from_iso3(self._country_config.iso3)
+        if self._iso2 is None:
+            raise KeyError(
```
turnerm, line 67: This file raises a lot of different errors for various situations, I would try to make sure those are covered by tests

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
-def _download_fewsnet_country(
-    date_pub: date,
-    iso2: str,
-    output_dir: Path,
-    use_cache: bool,
-) -> Optional[Path]:
-    """
-    Download fewsnet data that covers the iso2 country.
+        if self._country_config.fewsnet is None:
+            raise KeyError(
```
turnerm, line 72: Maybe this should be an AttributeError instead?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
-    return country_data
+        file_path = self._get_raw_file_projection_period(
+            dir_path=dir_path, projection_period=projection_period
+        )
+        return gpd.read_file(file_path)
 
+    def _check_date_validity(self, pub_year, pub_month):
+        try:
+            pub_date = datetime.datetime(year=pub_year, month=pub_month, day=1)
+        except ValueError as err:
```
turnerm, line 201: I don't think the first and last case are covered by your tests

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
+
+        Should either cover the iso2 or region.
+        If exists, returns the dir path.
+        """
+        country_dir = self._get_raw_dir_date(
+            area=self._iso2, pub_year=pub_year, pub_month=pub_month
+        )
+        if country_dir.is_dir():
+            return country_dir
+        else:
```
turnerm, line 319: You can remove the else, that way the nesting is reduced a bit

src/aatoolbox/config/countryconfig.py
turnerm: perfect!!

src/aatoolbox/config/countryconfig.py
```diff
@@ -31,11 +31,53 @@ class CodABConfig(BaseModel):
...
+        valid_regionnames = values["region_name_code_mapping"].keys()
+        if v not in valid_regionnames:
+            raise ValueError(
+                f"Invalid region name: {v}. "
+                f"Should be one of {valid_regionnames}"
+            )
+        return v
+
+    @root_validator(pre=False)
+    def _set_region_code(cls, values) -> dict:
```
turnerm, line 67: This is so cool, love it!

docs/source/datasources/fewsnet.rst
```diff
@@ -0,0 +1,91 @@
+FEWS NET
```
turnerm, line 1: Amazing documentation!

turnerm (commented):
Looks really awesome Tinka, amazing work! Just some last nit-picky stuff, so sorry...

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
-        only the year and month part are used
-        this commonly refers to the month of the Current Situation period
-    iso2 : str
-        iso2 code for which the data should be downloaded
-    output_dir: Path
-        path of dir to which the zip content should be written
-    use_cache: bool
-        if True, don't download if output_dir already exists
+        self._region_name = self._country_config.fewsnet.region_name
+        self._region_code = self._country_config.fewsnet.region_code
```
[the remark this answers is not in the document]
Tinkaa: I dont understand what you mean by this.. I mean `self._region_name` doesnt exist yet right? Or do you mean that I should call them later on in the code with `self._country_config.fewsnet.region_name` instead of `self._region_name`?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +33,335 @@
...
+                clobber=clobber,
+            )
+        except ValueError:
+            try:
+                return self._download_region(
+                    pub_year=pub_year,
+                    pub_month=pub_month_str,
+                    clobber=clobber,
+                )
+            except ValueError:
```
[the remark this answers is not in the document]
Tinkaa: I feel stupid, but I dont understand what you mean. You mean not adding an additional message, but just have the message of the original error (i.e. no valid file found). 
If so, imo I think it is useful to have this extra message that this data just doesnt exist, as the invalid zip file is a little bit more vague and doesnt show there is no country nor regional data that clearly. 

Extended the error message a bit, but let me know if you dont agree

src/aatoolbox/datasources/fewsnet/fewsnet.py
Tinkaa: hahaha nesting here we are. Should come up with a meme about it

Tinkaa: But I find not nesting very hard, so can you double-check if this is what you meant?

src/aatoolbox/datasources/fewsnet/fewsnet.py
Tinkaa: Coverage gutters shows that they aren't, but they definitely are.. One uses a month number above 12 so it is invalid. And the last one has the year 2100, i.e. it is in the future.

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
turnerm: I mean the latter, call them later with `self._country_config.fewsnet.region_name`

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
turnerm: I was thinking that you could have all the info needed in the original zipfile message. But agree that it perhaps doesn't belong there. In that case what you have looks good, I would just be sure to include the full traceback, i.e. raise `from err`

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +32,353 @@
...
+        pub_month_str = f"{pub_month:02d}"
+        dir_path = self._find_raw_dir_date(
+            pub_year=pub_year, pub_month=pub_month_str
+        )
+        file_path = self._get_raw_file_projection_period(
+            dir_path=dir_path, projection_period=projection_period
+        )
+        return gpd.read_file(file_path)
+
+    def _check_date_validity(self, pub_year, pub_month):
```
turnerm, line 218: This can be a static method

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +32,353 @@
...
+            zip_filename=self._get_zip_filename(
+                area=area, pub_year=pub_year, pub_month=pub_month
+            ),
+            url=url,
+        )
+
+    def _get_raw_dir_date(self, area, pub_year, pub_month):
+        return self._raw_base_dir / f"{area}_{pub_year}{pub_month}"
+
+    def _get_zip_filename(self, area, pub_year, pub_month):
```
turnerm, line 321: staticmethod

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +32,353 @@
...
+        raise FileNotFoundError(
+            f"No data found for {pub_year}-{pub_month} covering "
+            "{self._country_config.iso3} or {self._region_name}. "
+            "Please make sure the data exists and is downloaded"
         )
-    else:
-        region_data = output_dir_region
-    return region_data
+
+    def _get_raw_file_projection_period(self, dir_path, projection_period):
```
turnerm, line 351: staticmethod

turnerm (commented):
Sorry, one more thing, just noticed some methods that should be static

src/aatoolbox/datasources/fewsnet/fewsnet.py
turnerm: Aha yes, I think it's best to avoid having the main body of a function / method nested at all, if possible. And in this case your second return is 2x nested.

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
Tinkaa: Hmm the last answer didnt work either (error that Literal cannot accept tuples). Now used an enum but think it is overcomplicating things a bit for what it should do. Is this the way you imagined to use enum here?

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
Tinkaa: Yeah I would keep them separate. As I would clearly separate the case were no country data is found but there is regional data vs there is no country data nor regional data

Def should have the `from err`. Thanks for catching, now added.

tests/datasources/test_fewsnet.py
```diff
@@ -0,0 +1,237 @@
...
+    """Mock iso2 to iso3 conversion."""
+    if request.node.get_closest_marker("nomockiso2"):
+        return
+    mocker.patch(
+        "aatoolbox.datasources.fewsnet.fewsnet.Country.get_iso2_from_iso3",
+        return_value=ISO2.upper(),
+    )
+
+
+@pytest.mark.nomockiso2
```
Tinkaa, line 27: @turnerm I am becoming a test nerd. But maybe marking this, is a bit too much. Do you have an opinon on this?

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +37,281 @@
...
+
+    def process(self, *args, **kwargs):
+        """
+        Process FEWS NET data.
+
+        Method not implemented.
+        """
+        logger.info("`process()` method not yet implemented for FEWS NET.")
+
+    def load(  # type: ignore
```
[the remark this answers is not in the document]
turnerm: Hm, what do you think about putting a warning message in `load` saying that in future versions it will return the processed data? Will the two be very different or do you think it will be relatively backward compatible?

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
turnerm: I think it looks good with enum! My only suggestion would be instead of doing `projperiods = ValidProjectionPeriods._value2member_map_` (aha where did you find this method) you can do:
```
projperiods =  [pp.value for pp in ValidProjectionPeriods]
```
and maybe even just merge it with the next line

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
Tinkaa: I think it would change slightly significantly (aha). As now for the older data there is one shape polygon per IPC phase. And I would envision to have one polygon per livelihood zone (which is also what the newer raw data looks like). However, when looking at it nationally it thus should still contain the same information. 

But I like the idea of a warning message so will do that

tests/datasources/test_fewsnet.py
turnerm: Omg no this is so cool! You are a testing wizard Tinka

src/aatoolbox/datasources/fewsnet/fewsnet.py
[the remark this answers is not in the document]
Tinkaa: Discussed bilaterally and will be improved with the custom config PR, so change after that is merged

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -27,223 +32,361 @@
...
-        e.g. "EA"
-    output_dir : Path
-        path to dir to which the data should be written
-    use_cache : bool
-        if True, don't download if output_dir already exists
+    def _download_region(
+        self,
+        pub_year: int,
+        pub_month: str,
+        clobber: bool,
```
caldwellst, line 274: `clobber` isn't propagated through. I'm assuming it's meant to go through to `_download()` and then `_download_zip()` which is decorated?

PR #86, opened 2022-03-29
Use unique temporary directories for tests

turnerm:
- Ensure that any data the tests create goes into a unique temporary directory
- I know I should have done this separately but I realized that `module_base_dir` is a dumb name and changed it to `datasource_base_dir`


Tinkaa (approved):
Looks good!

PR #87, opened 2022-03-29
Fixes to GeoBoundingBox

turnerm:
I had started implementing these in the GloFAS branch, since that is delayed I thought better make a separate PR. The changes are:

- Rounding now returns new instance of GoeBoundingBox
- Use `Decimal` type internally. This is to fix some rounding errors I found with the GloFAS coordinates, e.g. I was getting 1.1500000000000001 after the rounding step, which was getting passed to the API. Probably not a huge deal but better to fix this I think.
- Don't allow the user to pass North < South or East < West. @Tinkaa when I implemented this I had to change the IRI tests - was this intentional on your part, do we not want this restriction?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to
 
 - `DataSource` is now an abstract base class with required `download`, `process`
   and `load` methods
+- `GeoBoundingBox.round` returns `GeoBoundingBox` instance (instead of being in place)
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


Tinkaa (approved):
I love it! Thanks for taking care of this, was exactly having the same issues but didnt get to fixing them. 

>Don't allow the user to pass North < South or East < West. @Tinkaa when I implemented this I had to change the IRI tests - was this intentional on your part, do we not want this restriction?

No it is great. I was just too lazy to implement before, so was using the rasters `invert_coordinates()` and thus wanted to test that that was working.

caldwellst (approved):

turnerm:
I updated `GeoBoundingBox.round` to return a new instance instead of being in-place. This solves the problem of datasources modifying a gbb that a user passes in.

Tinkaa (approved):
Thanks! Looks good!

PR #89, opened 2022-03-31
Feature/iri geobb

Tinkaa:
Just removed inverting coordinates step from IRI script as that is now taken care of by the check in geobounding box


Tinkaa:
Did figure out a bug though, that currently if we have geobounding box with e.g. west=4 and east=10, not the correct data is downloaded. This is fixable in the API, but might not be needed if we restrict the west coordinate to always be negative, see #92

turnerm (approved):
Looks good to go!

PR #93, opened 2022-04-06
HDX debug and tests

### LINKED ISSUE #76: Print out list of resources for HDX API in debug mode
@Tinkaa ran into an issue where a resource appeared on an HDX page, but was not in the list of resources found by the API. It would be helpful to be able to see the API resources to be able to diagnose this in the future.

turnerm:
Fixes #76, and I also added tests for the HDX API

### DIFF
--- a/src/aatoolbox/utils/hdx_api.py
+++ b/src/aatoolbox/utils/hdx_api.py
@@ -39,6 +39,7 @@ def load_dataset_from_hdx(
     """
     logger.info(f"Querying HDX API for dataset {hdx_address}")
     resources = Dataset.read_from_hdx(hdx_address).get_resources()
+    logger.debug(f"Found the following resources: {resources}")
     for resource in resources:
         if resource["name"] == hdx_dataset_name:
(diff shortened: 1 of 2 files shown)


mcarans (approved):
This looks good to me. While you could mock the whole Dataset object, what you have done is more precise.

(It looks like you are using hdx_dataset_name to refer to the resource name and I have assumed that for your usage, that is intended.)

Incidentally, your function could be useful to others, so you are welcome to make a PR to add it directly to the HDX Python API library

turnerm:
Thanks so much for taking a look! 

> This looks good to me. While you could mock the whole Dataset object, what you have done is more precise.

Would there be an advantage to doing it this way? Could I mock out the `get_resources` return value in a simpler way?

> (It looks like you are using hdx_dataset_name to refer to the resource name and I have assumed that for your usage, that is intended.)

Ah yes, maybe I can rename it for consistency.

> Incidentally, your function could be useful to others, so you are welcome to make a PR to add it directly to the HDX Python API library

Oo great, I will add that to my todo list!

Issue #92, opened 2022-04-04
GeoBoundingBox - only allow negative west coordinates? 


Tinkaa:
Due to the recent changes we made to the GeoBoundingBox and rethinking IRI, I was thinking if we should restrict the longitude range of GeoBoundingBox to be -180 to 180. 

We already restrict that East>West, which essentially means that the longitude range cannot be on the scale from 0-360, which makes sense imo. 

However, we can still have an input of say west=4 and east=10. While this is valid, I think the most commonly used longitude range is -180 to 180. Sometimes you see two times 0-180 used, where it is specifically specified whether it is West or East. When we do this, i.e. allowing positive values for the west coordinate, we need to do extra checks of the coordinate range for some data sources (e.g. for the IRI maproom downloads). 

Imo opinion it thus makes sense to restrict the west coordinate to be negative, but can also see arguments that it is too restricting. So curious to hear your opinions

turnerm:
This is a really good point, I had not thought of that! 

My initial instinct would be to restrict GeoBoundingBox to the most standard coordinate system, i.e. -180 to 180 for lon and -90 to 90 for lat. Do you think that would cause a problem for anything?

Tinkaa:
Ah yeah good point that this more than just negative values but also restricting ranges, also for latitude. Would agree on restricting.

Tinkaa:
Solved in #94

PR #94, opened 2022-04-06
GeoBoundingBox limits

turnerm:
Addresses #92. 

I wanted to check with you: I've now made it that south also cannot be equal to north (and west not equal to east). For the limits I've allowed the bounds to be included. Do you think these are the right choices?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to
 
 - `DataSource` is now an abstract base class with required `download`, `process`
   and `load` methods
+- `GeoBoundingBox` input parameters changed from `north`, `south`, `east`, and `west`
+  to `lat_max`, `lat_min`, `lon_max`, `lon_min`
 - `GeoBoundingBox.round` returns `GeoBoundingBox` instance (instead of being in place)
 
 ### Removed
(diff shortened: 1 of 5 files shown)


turnerm:
@Tinkaa thanks so much for taking a look!

> I think we should allow North==South, this would imo just indicating only retrieving that one lon/lat. What is your reasoning for only allowing >?

I figured that if you have North == South you have a line, not a box, not sure what the use case could be. 

> to me the names north, south, west, east, are getting more and more confusing. As you could also interpret it as being on the west vs east hemisphere. Would it be an idea to instead use max_lat, min_lat, max_lon, min_lon? Maybe I am nitpicking too much here though.

That's a great idea, I will change them!

Tinkaa:
>I figured that if you have North == South you have a line, not a box, not sure what the use case could be.
Yes agree you have a line and dont kno a use case either. But theoretically I think we should allow downloading a line.

turnerm:
> Yes agree you have a line and dont kno a use case either. But theoretically I think we should allow downloading a line.

Alright, done!

I've also changed the parameter names as you suggested. My last question for you @Tinkaa is this: do you think the string representations still make sense in light of this change? By that I mean the `__repr__`  and `get_filename_repr` methods

Tinkaa:
Hmm things are always more complex than you think. 

But I think it still makes sense to have the two functionalities and cannot come up with better naming so yeah would say we leave those.

Tinkaa (approved):

PR #99, opened 2022-04-11
Method decorator

caldwellst:
Addresses #62

### DIFF
--- a/src/aatoolbox/utils/io.py
+++ b/src/aatoolbox/utils/io.py
@@ -84,13 +84,13 @@ def parse_yaml(filename: Union[str, Path]) -> dict:
 
 
 @wrapt.decorator
-def check_file_existence(wrapped: F, instance, *args, **kwargs) -> F:
+def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
     """
     Don't overwrite existing data.
 
     Avoid recreating data if it already exists and if clobber not
-    toggled by user. Can be used to wrap a function whose first parameter
-    is a filepath.
+    toggled by user. Used to wrap functions that accept filepath
+    as a keyword argument.
 
     Parameters
     ----------
@@ -112,7 +112,7 @@ def check_file_existence(wrapped: F, instance, *args, **kwargs) -> F:
     the decorated function.
 
     """
-    filepath = args[0]
+    filepath = kwargs.pop("filepath")
     if filepath.exists() and not kwargs.get("clobber", False):
         logger.info(
             f"File {filepath} exists and clobber set to False, "
@@ -120,4 +120,5 @@ def check_file_existence(wrapped: F, instance, *args, **kwargs) -> F:
         )
         return filepath
     else:
+
         return wrapped(*args, **kwargs)
--- a/src/aatoolbox/utils/io.py
+++ b/src/aatoolbox/utils/io.py
@@ -112,6 +112,14 @@ def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
     the decorated function.
     """
     filepath = kwargs.get("filepath")
+    if filepath is None:
+        raise KeyError(
+            (
+                "`filepath` must be passed as a keyword "
+                "argument for the `check_file_existence`"
+                " decorator to work."
+            )
+        )
     if filepath.exists() and not kwargs.get("clobber", False):
         logger.info(
             f"File {filepath} exists and clobber set to False, "
--- a/src/aatoolbox/utils/io.py
+++ b/src/aatoolbox/utils/io.py
@@ -95,8 +95,8 @@ def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
     Parameters
     ----------
     wrapped : function
-        The function to wrap. The first parameter of this function must
-        be the filepath (of type Path), and it can also have an optional
+        The function to wrap. The function must have "filepath" as
+        a keyword parameter, and it can also have an optional
         "clobber" boolean keyword parameter.
     instance :
         Object the wrapped function is bound to. Not used within, but
--- a/src/aatoolbox/utils/io.py
+++ b/src/aatoolbox/utils/io.py
@@ -110,9 +110,8 @@ def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
     -------
     If filepath exists, returns filepath. Otherwise, returns the result of
     the decorated function.
-
     """
-    filepath = kwargs.pop("filepath")
+    filepath = kwargs.get("filepath")
     if filepath.exists() and not kwargs.get("clobber", False):
         logger.info(
             f"File {filepath} exists and clobber set to False, "
@@ -120,5 +119,4 @@ def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
         )
         return filepath
     else:
-
         return wrapped(*args, **kwargs)
--- a/src/aatoolbox/utils/io.py
+++ b/src/aatoolbox/utils/io.py
@@ -2,16 +2,12 @@
 import logging
 import zipfile
 from pathlib import Path
-from typing import Any, Callable, TypeVar, Union
+from typing import Union
 
 import requests
-import wrapt
 import yaml
 
 logger = logging.getLogger(__name__)
-
-# For typing the decorator
-F = TypeVar("F", bound=Callable[..., Any])
 
 
 def download_url(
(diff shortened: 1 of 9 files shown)


src/aatoolbox/utils/io.py
```diff
@@ -82,35 +83,40 @@ def parse_yaml(filename: Union[str, Path]) -> dict:
...
-        if filepath.exists() and not kwargs.get("clobber", False):
-            logger.info(
-                f"File {filepath} exists and clobber set to False, "
-                f"using existing files"
-            )
-            return filepath
-        return func(filepath, *args, **kwargs)
-
-    return cast(F, wrapper)
+    filepath = kwargs.get("filepath")
```
Tinkaa, line 114: only thing is that this is now a kwarg. So if say you have
```
@check_file_existence
def _download(filepath):
  ....
``` 
and you call `_download(filepath)` instead of `_download(filepath=filepath)`, it gives an error. Which is not bad, but maybe we could return an easier understandable error message cause now it is `AttributeError: 'NoneType' object has no attribute 'exists'`

Tinkaa (approved):
Looks good. As far as I understand, wrapt is just a library that makes decorators more transparent and thus less prone to hidden errors. Is that correct?

src/aatoolbox/utils/io.py
caldwellst: Yeah, thought about that. Previously was using `kwargs.pop()` which provides clear error message but of course, we actually needed `"filepath"` still in the dictionary when we pass it later on to `wrapped()` (as an aside, took me days to notice that was happening and only with Monica's help eventually realized hahaha). Since `_download()` is internal, wasn't sure if we should do error checking here or just assume we construct in the correct way so that `download()` would always call it with a keyword argument?

caldwellst:
The main benefit of wrapt is not just the transparency, but also that it allows you to write method decorators that work on functions, static methods, class methods, classes, and instance methods, without having to change anything. Without it, you have to do workarounds since class methods for instance pass `self` as the first argument, so it breaks a decorator that is designed for functions or static methods.

src/aatoolbox/utils/io.py
caldwellst: Addressed!

turnerm: Actually, why not use a `try` here, catch the key error, and re-raise with the improved error message?

src/aatoolbox/utils/io.py
```diff
@@ -82,35 +83,48 @@ def parse_yaml(filename: Union[str, Path]) -> dict:
     return config
 
 
-def check_file_existence(func: F) -> F:
+@wrapt.decorator
+def check_file_existence(wrapped: F, instance, args, kwargs) -> F:
```
turnerm, line 87: Is it possible to type these arguments somehow? Like should args be a list and kwargs a dict? And could instance be `DataSrouce` (would it then accept child classes?)

src/aatoolbox/utils/io.py
```diff
@@ -82,35 +83,48 @@ def parse_yaml(filename: Union[str, Path]) -> dict:
...
+                "`filepath` must be passed as a keyword "
+                "argument for the `check_file_existence`"
+                " decorator to work."
             )
-            return filepath
-        return func(filepath, *args, **kwargs)
-
-    return cast(F, wrapper)
+        )
+    if filepath.exists() and not kwargs.get("clobber", False):
```
turnerm, line 123: I think this part here is not covered by any tests, would it be possible to add some? Sorry Seth, I know I'm terrible because I'm the one who wrote the function in the first place and didn't make tests! Only if you have time.

turnerm (commented):
Hey Seth, this looks amazing, the `wrapt` package was a great find! 

My only major comment would be -- do you think you could go into some of the data sources and change the decorated download functions to instance methods? That way we are really making good use of this improvement, and I think it'll be very nice to see it in action.

src/aatoolbox/utils/io.py
caldwellst: I am trying to figure out, what IS self for passing? I think we are mainly concerned with decorating instance methods, right, rather than class methods. How do you type hint self? It isn't a class I think, right? Like somehow it's different? Or if it's self do we just try `Type[DataSource]` to specify that it's [a subclass of DataSource].(https://stackoverflow.com/questions/46092104/subclass-in-type-hinting)

src/aatoolbox/utils/io.py
caldwellst: Not sure exactly what you mean. `kwargs.get()` doesn't raise an error message, it just returns `None`, hence the need to check. Don't want to use `.pop()`, which does raise an error message, because we want to keep the `kwarg` in the dict for passing to `wrapped`.

src/aatoolbox/utils/io.py
caldwellst: Figured out the typing, I think we can directly pass the abstract class and require it. I think it is working (according to tests). See the thread [here](https://stackoverflow.com/questions/38791739/function-annotation-for-subclasses-of-abstract-class).

caldwellst:
I have updated the mocking, just figured out I had to of course reference the class of the instance method in the mocker patch. Still don't understand it but we can keep that for next week. I have also put `DataSource` in the type hints for the decorator. However, this created issues with circular imports within `io.py`, so I have put the decorator into its own file. We can change file name if we want, just went simplest option.

src/aatoolbox/utils/io.py
caldwellst: Will add in as hopefully final step!

src/aatoolbox/utils/io.py
turnerm: You can just do:
```python
try:
    filepath = kwargs["filepath"]
except KeyError as err:
   raise KeyError("message") from err
```

turnerm:
> However, this created issues with circular imports within io.py, 

Oh no, what were the issues? We can also leave it untyped (or is there like a general type that would cover it?) Maybe DataSource is too restrictive anyway. Sorry to make you go in circles!

caldwellst:
The issue was just specifically circular imports. Because to type it as DataSource in the `io.py` file required importing from `datasource.py`, which itself imports from `countryconfig.py` which itself imported from `io.py`. It was just that I needed it in a separate file to remove the import loop. I think is okay that way, but let me know. To make it more flexible, so that it could be a static method or function if we want, I have added `Optional[DataSource]`, which I think is optimal. What do y'all think?

Tinkaa:
Looks great @caldwellst !

caldwellst:
@turnerm @Tinkaa Okay, last thing I did was just add in the tests. I didn't use `mocker`, I just used `tmp_path.touch()` to create the fake file and test `check_file_existence()`. We can always refactor, but at least covering the main functionality of the decorator now for that sweet, sweet  100% coverage.

tests/test_check_file_existence.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_fp_exists_no_clobber(tmp_path):
+    """Test that filepath returned if Path exists and clobber False."""
+    tmp_path.touch()
+    output_filepath = downloader(filepath=tmp_path, clobber=False)
+    assert output_filepath == tmp_path
+
+
+def test_fp_exists_clobber(tmp_path):
+    """Test that filepath returned if Path exists and clobber False."""
```
Tinkaa, line 22: Should be "Test that function returned if Path exists and clobber True?

tests/test_check_file_existence.py
```diff
@@ -0,0 +1,37 @@
...
+    tmp_path.touch()
+    output_filepath = downloader(filepath=tmp_path, clobber=True)
+    assert output_filepath == "a"
+
+
+def test_fp_not_exists(tmp_path):
+    """Test that filepath returned if Path does not exist."""
```
Tinkaa, line 29: Should be "Test that function returned if Path does not exist."?

Tinkaa (approved):
Looks good! I dont think you need mocking in this case. 

Should we also do testing if the function is within a class or we can trust that `wrapt` does that correctly?

src/aatoolbox/utils/io.py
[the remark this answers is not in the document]
caldwellst: This code is no longer in the PR, it's done now using `kwargs["filepath"]` and try/except to explicitly catch the error with an informative message.

caldwellst:
I would say we can trust that `wrapt` does it correctly, and it will be tested by all of our instance methods in the various process/download methods, so we can avoid that? Will merge if @turnerm also approves.

turnerm (approved):
Yes looks amazing, thanks for implementing this!

Issue #91, opened 2022-04-04
Add NDVI


turnerm:
Migrate contents of [this script](https://github.com/OCHA-DAP/pa-anticipatory-action/blob/develop/src/indicators/drought/ndvi.py)

PR #100, opened 2022-04-13
Add NDVI class, address #91

caldwellst:
Okay, so the code for the NDVI class is there and ready to review. The only thing left is to add in tests. However, I am putting this out now because I will need to work with Tinka to better understand mocking and test setup, but there is sufficient code (a lot of it) for review already.


src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    end_date : Union[date, str, Tuple[int], None]
+        End date. Can be passed as a ``datetime.date``
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    processed_file_suffix : str
+        Suffix for processed file.
```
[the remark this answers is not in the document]
Tinkaa: Or actually, instead of having the default an empty string, I would set it to `None`

Tinkaa (approved):
Looking great! 

Some things I would change regarding setting variables and file structure. But tested it and it works perfectly. 

Mainly looked at the workings on the code, maybe @turnerm should come in at some point as well with her master-ness on design principles etc.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+import pandas as pd
+import regex as re
+import rioxarray  # noqa: F401
+import xarray as xr
+from typing_extensions import Literal
+
+import aatoolbox.utils.raster  # noqa: F401
+from aatoolbox.config.countryconfig import CountryConfig
+from aatoolbox.datasources.datasource import DataSource
+from aatoolbox.utils._dates import (
```
turnerm, line 52: `_dates` and all the functions below should be public, since you're importing them

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        url = self._get_url(filename=filename)
+        try:
+            resp = urlopen(url)
+        except HTTPError:
+            year, dekad = self._fp_year_dekad(filepath)
+            logger.error(
+                f"No NDVI data available for "
+                f"dekad {dekad} of {year}, skipping."
+            )
+            pass
```
turnerm, line 723: This is not needed

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+
+        # open file within memory
+        zf = ZipFile(BytesIO(resp.read()))
```
turnerm, line 726: It seems that you can reach this point with `resp` being undefined, because if you get an HTTP error you just log it but don't exit the method

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+
+
+class _UsgsNdvi(DataSource):
+    """Base class to retrieve USGS NDVI data.
+
+    Parameters
+    ----------
+    country_config : CountryConfig
+        Country configuration
+    coverage_area : str
```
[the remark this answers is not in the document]
turnerm: Yes agreed, Tinka did an amazing job with this, see [here](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/84/files#diff-8a6addc4910e8fe8b3540f3c9a2b38b998eb800953188c856e2aa8092fbf1198R46). This way you'll avoid having variables like `loc_region_types` throughout the code as they should really be constants / in the config.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        Area of coverage
+    data_type : str
+        Data type
+    start_date : Union[date, str, Tuple[int], None]
```
turnerm, line 92: Can there be defaults for start and end date, e.g. spanning the full date range?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        end_date: Union[date, str, Tuple[int], None] = None,
+        processed_file_suffix: str = "",
+    ):
+        super().__init__(
+            country_config=country_config,
+            datasource_base_dir="usgs_ndvi",
+            is_public=True,
+            is_global_raw=True,
+        )
+        self.coverage_area = coverage_area
```
turnerm, line 121: This can be private I think

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        valid_types = _VALID_TYPES.__args__  # type: ignore
+        if data_type not in valid_types:
+            raise ValueError(
+                "`data_type` is not a valid string value. "
+                f"It must be one of {*valid_types,}"
+            )
+        self._data_type = data_type
+
+        # url code and file suffix
+        loc_data_types = {
```
turnerm, line 134: To me it makes more sense to not have a dictionary for this, but rather pass all the parameters (i.e. `data_type_url`, `data_type_suffix` ) directly to `__init__`. The reason for this is if you add another data type, then you would need to update the dictionary that's buried in the base class __init__. If you make them all input parameters, you would know right away that you need to provide this info. Hope that makes sense.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        if value not in valid_args:
+            raise ValueError(
+                "`coverage_area` is not a valid string value. "
+                f"It must be one of '{*valid_args,}'."
+            )
+        self._coverage_area = value
+
+        # also adjust relevant URL and file attributes
+        # when coverage area is set
+        loc_region_types = {
```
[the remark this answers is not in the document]
turnerm: Agreed!

turnerm (commented):
I need to hit pause for now, but I just wanted to say Seth, what an incredible amount of work! Looking forward to digging in more next week :)

turnerm (commented):
Alright, I've looked through everything now ;) Looks really great Seth! 

Last couple of things:
- Are you sure you committed in the correct venv? I think there are some pre-commit issues with mypy.
- The `usgs_ndvi` module is quite long, if after the suggested changes it stays that way, maybe you could consider splitting it up, like moving the child classes to their own module or something

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            "percent_median": ("percentofmedian", "pct"),
+            "median_anomaly": ("mediananomaly", "stmdn"),
+            "difference": ("differencepreviousyear", "dif"),
+        }
+        data_type_values = loc_data_types[data_type]
+        self._data_type_url = data_type_values[0]
+        self._data_type_suffix = data_type_values[1]
+
+        # set dates for data download and processing
+        if start_date is None:
```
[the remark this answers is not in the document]
caldwellst: Yes, good idea!!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            # earliest dekad of data is 2002-19
+            self._start_year, self._start_dekad = 2002, 19
+        elif isinstance(start_date, (str, date)):
+            self._start_year, self._start_dekad = _date_to_dekad(start_date)
+        else:
+            self._start_year, self._start_dekad = cast(list, start_date)
```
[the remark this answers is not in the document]
caldwellst: Was using `cast` to specifically specify the typing, now just at 1 location, essentially ensuring `mypy` passes there (knowing that we aren't checking for length on one of the other types). Just using `list()` produces typing errors, so have reverted to being even more explicit here to satisfy `mypy` and ensure everything is working correctly.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+    processed_file_suffix : str
+        Suffix for processed file.
+    """
+
+    def __init__(
+        self,
+        country_config: CountryConfig,
+        coverage_area: _VALID_AREAS,
+        data_type: _VALID_TYPES,
+        start_date: Union[date, str, Tuple[int], None] = None,
```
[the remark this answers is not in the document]
caldwellst: Fair question. However, since the tuple format is for dekads, I thought providing the option to pass in more user friendly dates would be easier. For instance, If you want all data from July 1 onward, I would find it much easier to pass `2020-07-01` as a string than figuring out what the year and dekad is. I've actually gotten all the `mypy` stuff running with checks, so let me know what you think.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+
+        if end_date is None:
+            self._end_year, self._end_dekad = _date_to_dekad(date.today())
+        elif isinstance(start_date, (str, date)):
+            self._end_year, self._end_dekad = _date_to_dekad(end_date)
+        else:
+            self._end_year, self._end_dekad = cast(list, end_date)
+
+        # set processed file suffix
+        self.processed_file_suffix = processed_file_suffix
```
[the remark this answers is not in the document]
caldwellst: This is a public property. Added it because needed a way for users to specify the file suffix so they can `process()` to different levels of aggregation.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            # get dates that have already been processed
+            processed_dates = df[["year", "dekad"]].values.tolist()
+            if clobber:
+                if cols_same:
+                    # remove processed dates from file
+                    # so they can be reprocessed
+                    keep_rows = [
+                        d not in process_dates for d in processed_dates
+                    ]
+                    df = df[keep_rows]
```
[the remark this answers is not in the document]
caldwellst: I think `.loc` is only used if you [want a selection of rows and columns](https://pandas.pydata.org/docs/getting_started/intro_tutorials/03_subset_data.html).

src/aatoolbox/utils/_dates.py
```diff
@@ -0,0 +1,98 @@
+"""Functions for dealing with dates."""
```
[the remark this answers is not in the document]
caldwellst: I was thinking it should as a private module, based on the same link ha! In the accepted answer it says:

> `_...` is often used as 'internal' name. For example, modules starting with `_` shouldn't be used directly; similarly, methods with `_` are supposedly-private and so on. It's just a convention but you should respect it.

So was putting this as a private module. However, see the answer below that you're referring to, so can put this just as `dates.py` because it seems actually unnecessary.

src/aatoolbox/utils/_dates.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Dekad computed from date. This
+    is based on the USGS (and relatively
+    common) dekadal definition of the
+    1st and 2nd dekad of a month being
+    the first 10 day periods, and the 3rd
+    dekad being the remaining days within
+    that month.
+    """
+    if isinstance(date_obj, str):
```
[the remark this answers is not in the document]
caldwellst: Smart, okay, have adjusted!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
+"""Class to download and process USGS FEWS NET NDVI data.
+
+Data is downloaded from the `USGS FEWS NET data portal
+<https://earlywarning.usgs.gov/fews>`_. Data is
+generated from eMODIS AQUA, with full methodological
+details available from the various data pages
+on the data portal. The available areas of coverage are:
```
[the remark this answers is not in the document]
caldwellst: What do you mean under `Documentation`?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+    ----------
+    country_config : CountryConfig
+        Country configuration
+    coverage_area : str
+        Area of coverage
+    data_type : str
+        Data type
+    start_date : Union[date, str, Tuple[int], None]
+        Start date. Can be passed as a ``datetime.date``
+        object and the relevant dekad will be determined,
```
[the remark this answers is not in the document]
caldwellst: You are definitely right!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
Tinkaa: If you go to the links that ar elinked here, you can click on the tab Documentation to see the actual methodological details. Is that correct? Cause I would state that clearly then as I was now confused where I should find the methodological details based ont  "the various data pages on the data portal"

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+    """Base class to retrieve USGS NDVI data.
+
+    Parameters
+    ----------
+    country_config : CountryConfig
+        Country configuration
+    coverage_area : str
+        Area of coverage
+    data_type : str
+        Data type
```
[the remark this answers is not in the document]
caldwellst: Still a good idea to be consistent and clear, have changed.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: Yes makes sense!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        >>>
+        >>> # Retrieve admin 2 boundaries for Nepal
+        >>> country_config = create_country_config(iso3="npl")
+        >>> codab = CodAB(country_config=country_config)
+        >>> npl_admin2 = codab.load(admin_level=2)
+        >>>
+        >>> # setup NDVI
+        >>> UsgsNdviSmoothed(
+        ...     country_config=country_config,
+        ...     coverage_area="west_africa",
```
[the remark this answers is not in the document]
caldwellst: hahahaha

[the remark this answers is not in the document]
caldwellst: Pangaea forever!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+dekadal temporally smooth NDVI data for West Africa is available at
+`this link
+<https://edcintl.cr.usgs.gov/downloads/sciweb1/shared/fews/web/africa/west/dekadal/emodis/ndvi_c6/temporallysmoothedndvi/downloads/monthly/>`_
+
+The products include temporally smoothed NDVI, median anomaly,
+difference from the previous year, and median anomaly
+presented as a percentile.
+
+Data by USGS is published quickly after the dekad.
+After about 1 month this data is updated with temporal smoothing
```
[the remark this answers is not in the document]
caldwellst: Yes, thought about that last week and just didn't have brainspace to address yet. I think will need system to set clobber separately for recent, so that by default it re-downloads files in the last 1 month each time. It's not indicated in the file in any way, sadly, as they are just `.tif` files with a georeferencing `.tfw` file. Do you have any potential ideas on this? I'm still trying to think through.

If we just update the past 1 month, I think that's a safe bet. However, maybe I come back after a year and want to update. Could we myabe pass in `clobber_since` as an argument, that defaults to `None`, but when passed only clobbers files after that date?

src/aatoolbox/utils/_dates.py
```diff
@@ -0,0 +1,98 @@
...
+    """Is year1/dekad1 less than year2/dekad2.
+
+    Compare two pairs of years and dekads,
+    that the first pair are less than the
+    second pair.
+    """
+    return y1 < y2 or ((y1 == y2) and (d1 < d2))
+
+
+def _compare_dekads_lte(y1: int, d1: int, y2: int, d2: int):
```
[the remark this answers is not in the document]
caldwellst: It's not. But figured I would just put in with the other 2 in case useful in the future, since there could be use cases for >= and <= operations in the same way. for instance if we do a `clobber_since` argument or something like discussed above.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        are passed on to ``aat.computer_raster_stats()``.
+        ``self.processed_file_suffix`` is used to define
+        the unique processed files, and should be changed
+        if the user wants to analyze across multiple files.
+
+        Parameters
+        ----------
+        gdf : geopandas.GeoDataFrame
+            GeoDataFrame with row per area for stats computation.
+            If ``pd.DataFrame`` is passed, geometry column must
```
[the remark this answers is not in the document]
caldwellst: Yeah, it tracks your specified geometry column. It's just that `geometry` is the default!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            Dekad
+
+        Returns
+        -------
+        str
+            File path prefix for .zip file at URL and
+            for .tif files stored within the .zip
+        """
+        file_name = (
+            f"{self._region_prefix}{year-2000:02}"
```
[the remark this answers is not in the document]
caldwellst: Yeah, probably best to change. Was keeping the naming identical to what is stored on USGS website (in the zip file), but can adjust. Will take some adjustments to the file loading/pathing functions.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+from io import BytesIO
+from pathlib import Path
+from typing import List, Tuple, Union, cast
+from urllib.error import HTTPError
+from urllib.request import urlopen
+from zipfile import ZipFile
+
+import geopandas as gpd
+import pandas as pd
+import regex as re
```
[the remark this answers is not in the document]
caldwellst: I hadn't realized there were two...

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
turnerm: Oops you're right sorry!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
turnerm: But doesn't indexing a dataframe directly with a list like this refer to the columns, rather than the rows? Sorry I might be being totally dense here

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Okay clear. Still wondering about their naming in regards to the overall API. https://stackoverflow.com/questions/3602110/python-private-module-in-a-package

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Oooh whoops, I had thought `pass` exited the function, but instead have put in an empty return now.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: That's very clear, thanks!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: Removing anyway based on other comments about the suffix and instead based on aggregation.

src/aatoolbox/utils/_dates.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    if isinstance(date_obj, str):
+        date_obj = date.fromisoformat(date_obj)
+
+    year = date_obj.year
+    dekad = (date_obj.day // 10) + ((date_obj.month - 1) * 3) + 1
+    return year, dekad
+
+
+def _compare_dekads_lt(y1: int, d1: int, y2: int, d2: int):
```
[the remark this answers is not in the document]
caldwellst: Yeah, agreed, I like the class idea you said earlier and the lnk you sent very interesting!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: Won't let me resolve this lol.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        ...     end_date=[2020, 3]
+        ... )
+        >>> UsgsNdviSmoothed.download()
+        >>> UsgsNdviSmoothed.process()
+        >>>
+        >>> # process for admin
+        >>> UsgsNdviSmoothed.processed_file_suffix = "adm1"
+        >>> UsgsNdviSmoothed.process()
+        """
+        processed_path = self._get_processed_path()
```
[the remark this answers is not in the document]
caldwellst: Why not? Not sure I understand.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            d1=self._start_dekad,
+            y2=self._end_year,
+            d2=self._end_dekad,
+        )
+
+        # check to see if file exists and remove
+        # if clobber or check dates already processed
+        # if not
+        if processed_path.is_file():
+            df = self.load()
```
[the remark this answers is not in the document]
caldwellst: Yeah, good idea!!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        Returns
+        -------
+        List
+            List of year and dekad lists for files already
+            in the system.
+        """
+        regex = fr"{self._region_prefix}[0-9]{{4}}{self._data_type_suffix}.tif"
+        downloaded_dekads = [
+            self._fp_year_dekad(filename)
+            for filename in self._raw_base_dir
```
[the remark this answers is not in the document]
caldwellst: Yes, you're right, needed to iterate over the files within!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+                            "as at a different disaggregation)."
+                        )
+                    )
+        else:
+            # empty data frame for concatting later
+            df = pd.DataFrame()
+
+        # process data for necessary dates
+        data = [df]
+        for d in process_dates:
```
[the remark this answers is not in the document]
caldwellst: I did but then it creates errors/confusion with `datetime.date`.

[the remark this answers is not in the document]
caldwellst: have put as `process_date`

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+        ...     country_config=country_config,
+        ...     coverage_area="west_africa",
+        ...     start_date=[2020, 1],
+        ...     end_date=[2020, 3]
+        ... )
+        >>> UsgsNdviSmoothed.download()
+        """
+        for year in range(self._start_year, self._end_year + 1):
+            for dekad in range(1, 37):
+                if year == self._start_year and dekad < self._start_dekad:
```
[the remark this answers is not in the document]
caldwellst: ...How did I not think of that. Jesus.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            # check that the processed file has the same analyzed
+            # indicators and column for aggregation as passed
+            # to process()
+            cols = kwargs.get(
+                "stats_list", ["mean", "std", "min", "max", "sum", "count"]
+            )
+            percentile_list = kwargs.get("percentile_list")
+            if percentile_list is not None:
+                for percent in percentile_list:
+                    cols.append(f"{percent}quant")
```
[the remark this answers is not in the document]
caldwellst: I looked into this, but couldn't figure out a convenient way to do a list comprehension on an existing list (since we are coming in with `cols` already filled out from above. Any suggestions?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
Tinkaa: Aaah okay, if it is identical to what is being used on the USGS, then maybe stick to that.. Could we document it somewhere then though?

src/aatoolbox/utils/_dates.py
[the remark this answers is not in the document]
caldwellst: Will leave for the moment and hope we can address all of this in a dekad class.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Since we want to eventually put into a class, will resolve for now I don't think is huge deal at the moment.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,924 @@
...
+            d2=self._end_dekad,
+        )
+
+        # check to see if file exists and remove
+        # if clobber or check dates already processed
+        # if not
+        if processed_path.is_file():
+            df = self.load()
+
+            # check that the processed file has the same analyzed
```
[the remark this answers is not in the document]
caldwellst: So, even before, the `feature_col` was checked to make sure that the files matches with the `feature_col` for aggregation and the stats being calculated. However, I agree with y'alls comments that we should use `feature_col` for column identification, and have changed for that. I see the advantage of different files, just don't think that having a timestamp solves the problem of overwriting/metadata, since it doesn't help match up the current call with the processed file. I couldn't think of additional metadata to add, but having the `feature_col` seems pretty sufficient I think?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
Tinkaa: Pff horrible, this should be USGS responsibility. 
Buttt would just checking if the last modified date of the url is later than the download date? 

I did a bit of googling and the accepted answer [here](https://stackoverflow.com/questions/29314287/python-requests-download-only-if-newer) seems to be working. Where the returned `url_time` is the same as that indicated under Last Modified [on their website](https://edcintl.cr.usgs.gov/downloads/sciweb1/shared/fews/web/africa/east/dekadal/emodis/ndvi_c6/percentofmedian/downloads/dekadal/)

The date that is returned as `file_time` is the date the file was downloaded and thus not the one it was modified, but if we just check that that one is larger than `url_time` it shouldn't really matter. 

The computation of the times doesn't seem to take super long so hopefully wouldn't slow it down too much. 

I dont know if it is a solid approach and we should at least test it with a file that will get updated soon. But what do you think?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
Tinkaa: Yeah I would agree that that is easier while we prob still want the option of the dekad number as well. So would leave agree to leave as it is.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
Tinkaa: hmmm yeah agree a lot is solved with adding the `feature_col`. Now I think the only case when it is overwritten is when using different stats, is that correct? 

One idea to prevent that is that we could have a separate file per stat? and then with loading combine them? Cause the risk I see now is that e.g. one person calls the function only computing `mean`. Someone else with only `sum`. And that would mean (ha pun) that suddenly the `mean` results disappear and would be hard to traceback why. What you think of that?

src/aatoolbox/config/countries/bfa.yaml
```diff
@@ -3,3 +3,7 @@ codab:
   hdx_dataset_name: bfa_adm_igb_20200323_SHP.zip
   layer_base_name: bfa_admbnda_adm{admin_level}_igb_20200323.shp
   admin_level_max: 3
+fewsnet:
+  region_name: west-africa
+usgs_ndvi:
+  region_name: west-africa
```
Tinkaa, line 9: Ha tests are gonna save you here. But now getting an error as this par is named `region_name` while countryconfig is expecting `area_name`

src/aatoolbox/config/countryconfig.py
```diff
@@ -26,33 +26,160 @@ class CodABConfig(BaseModel):
...
+            raise ValueError(
+                f"Invalid area name: {v}. "
+                f"Should be one of {', '.join(valid_area_names)}"
+            )
+        return v
+
+    @root_validator(pre=False, skip_on_failure=True)
+    def _set_area_codes(cls, values) -> dict:
+        """Set NDVI url and prefix from area."""
+        area_values = values["area_name_mapping"]["area_name"]
```
Tinkaa, line 120: getting an error here. Should be `area_values = values["area_name_mapping"][values["area_name"]]

src/aatoolbox/config/countryconfig.py
```diff
@@ -81,12 +81,54 @@ def _set_region_code(cls, values) -> dict:
...
+        values["area_url"], values["area_prefix"] = area_values
```
Tinkaa, line 121: very small, but would combine this to one line, i.e. uneccessary to create `area_values`

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -26,56 +28,44 @@
...
-    _date_to_dekad,
     _dekad_to_date,
     _expand_dekads,
+    _get_dekadal_date,
 )
 
-# TODO: use check_file_existence once the fix is merged
-# that allows it to work on class methods (PR is open)
-# from aatoolbox.utils.io import check_file_existence
+# from aatoolbox.utils.check_file_existence import check_file_existence
```
Tinkaa, line 63: why is this and the decorator at `_download` commented out?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -85,117 +75,85 @@ class _UsgsNdvi(DataSource):
...
-            "smoothed": ("temporallysmoothedndvi", ""),
-            "percent_median": ("percentofmedian", "pct"),
-            "median_anomaly": ("mediananomaly", "stmdn"),
-            "difference": ("differencepreviousyear", "dif"),
-        }
-        data_type_values = loc_data_types[data_type]
-        self._data_type_url = data_type_values[0]
-        self._data_type_suffix = data_type_values[1]
+        self._area_url = self._country_config.usgs_ndvi.area_url
+        self._area_prefix = self._country_config.usgs_ndvi.area_prefix
```
Tinkaa, line 126: sorry, this wasnt updated yet in the FN pr where I guess you got this from, but now it is (looks like you didn't merge the latest `develop` yet?). Instead of `self._country_config.usgs_ndvi.area_url` you can now use `self._datasource_config.area_url` . And then here, you don't have to assign it anymore to a new var `self._area_url`, but just use the `self._datasource_config.area_url` when you need it. Reason we were assigning it to a new var before is because otherwise mypy would complain

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -85,117 +75,85 @@ class _UsgsNdvi(DataSource):
...
 
-        if end_date is None:
-            self._end_year, self._end_dekad = _date_to_dekad(date.today())
-        elif isinstance(start_date, (str, date)):
-            self._end_year, self._end_dekad = _date_to_dekad(end_date)
-        else:
-            self._end_year, self._end_dekad = cast(list, end_date)
+        self._end_year, self._end_dekad = _get_dekadal_date(
+            input_date=end_date, default_date=date.today()
+        )
```
Tinkaa, line 140: nice!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -314,30 +274,35 @@ def process(
...
+        >>> bfa_ndvi.process(
+        ...     gdf=bfa_admin2,
+        ...     feature_col="ADM2_FR"
+        ... )
         >>>
-        >>> # process for admin
-        >>> UsgsNdviSmoothed.processed_file_suffix = "adm1"
-        >>> UsgsNdviSmoothed.process()
+        >>> # process for admin1
+        >>> bfa_ndvi.processed_file_suffix = "adm1"
```
Tinkaa, line 296: Is this still needed, as the `feature_col` name is now automatically appended to the filename?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -348,78 +313,32 @@ def process(
...
         # if not
         if processed_path.is_file():
-            df = self.load()
-
-            # check that the processed file has the same analyzed
-            # indicators and column for aggregation as passed
-            # to process()
-            cols = kwargs.get(
-                "stats_list", ["mean", "std", "min", "max", "sum", "count"]
+            dates_to_process, df = self._determine_process_dates(
```
Tinkaa, line 316: confusing that the output name of this is the same. Maybe change the output name of `_expand_dekads` to smth like `included_dates`?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -455,23 +382,25 @@ def load(self) -> pd.DataFrame:
...
         ... )
-        >>> UsgsNdviSmoothed.download()
-        >>> UsgsNdviSmoothed.process()
-        >>> UsgsNdviSmoothed.load()
+        >>> bfa_ndvi.download()
+        >>> bfa_ndvi.process(
+        ...    gdf=bfa_admin2,
+        ...    feature_col="ADM2_FR"
+        )
+        >>> bfa_ndvi.load()
```
Tinkaa, line 401: needs the `feature_col` arg for `load()`

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -600,40 +688,27 @@ def _get_raw_path(self, year: int, dekad: int) -> Path:
             4-digit year
         dekad : int
             Dekad
+        local : bool
```
Tinkaa, line 691: nice solution!

src/aatoolbox/datasources/usgs/usgs_ndvi_children.py
```diff
@@ -0,0 +1,226 @@
...
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    end_date : Union[date, str, Tuple[int, int], None]
+        End date. Can be passed as a ``datetime.date``
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    processed_file_suffix : str
+        Suffix for processed file.
```
Tinkaa, line 74: I would incline towards also adding examples here, but maybe it is an overkill. What would you say?

src/aatoolbox/datasources/usgs/usgs_ndvi_children.py
```diff
@@ -0,0 +1,226 @@
...
+        Start date. Can be passed as a ``datetime.date``
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    end_date : Union[date, str, Tuple[int, int], None]
+        End date. Can be passed as a ``datetime.date``
+        object and the relevant dekad will be determined,
+        as a date string in ISO8601 format, or as a
+        year-dekad tuple, i.e. (2020, 1).
+    processed_file_suffix : str
```
Tinkaa, line 73: Add that this defaults to None

src/aatoolbox/datasources/usgs/usgs_ndvi_children.py
```diff
@@ -0,0 +1,226 @@
...
+        Suffix for processed file.
+    """
+
+    def __init__(
+        self,
+        country_config: CountryConfig,
+        start_date: Union[date, str, Tuple[int, int], None] = None,
+        end_date: Union[date, str, Tuple[int, int], None] = None,
+        processed_file_suffix: str = "",
```
Tinkaa, line 82: initiatie with None instead of ""?

Tinkaa (approved):

Tinkaa:
Looking great @caldwellst , so nice to see it being even better than before!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: That sounds like a really great approach, and makes loads of sense. I will look into that link you sent, thanks, that's brilliant!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: That's another brilliant idea, and solves the rest of the issues. Will have a think through the cleanest way to implement this, and put something in to match.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: I am having trouble at the moment rebasing to develop even though these functions are now in there. Not sure why. Trying to sort.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: NOPE!

src/aatoolbox/datasources/usgs/usgs_ndvi_children.py
caldwellst: No, i think you're definitely right!

src/aatoolbox/config/countries/bfa.yaml
caldwellst: Ha yes, haven't tested this stuff yet in new, will do. Thanks for catching up!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
turnerm: Really love the idea @Tinkaa!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
turnerm: I think you should just be able to do `existing_list += [comprehension]`

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: Done and implemented! Had to think through how to also adjust for this in the `process()` method. Let me know what y'all think.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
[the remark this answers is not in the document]
caldwellst: Oh wow. So simple and clean, of course you can. Great idea!!

docs/source/datasources/usgs_ndvi.rst
```diff
@@ -0,0 +1,131 @@
...
+    ndvi_smooth.download()
+
+Since each NDVI file is downloaded as a .tif, we can load them in individually as
+rasters. We can do this for a specific date. This time we will specify the date
+as a year and dekad tuple, although specify as a `datetime.date` or string as
+above is equally appropriate.
+
+.. code-block:: python
+
+    ndvi_2021_01 = ndvi_smooth.load_raster(date=(2021, 01))
```
Tinkaa, line 104: is this trailing 0 necessary? If so, maybe dont leave that to the user but add that in the code (can use `:02d`)

docs/source/datasources/usgs_ndvi.rst
```diff
@@ -0,0 +1,131 @@
...
+
+If time series analysis is desired on the NDVI data, the user can manually do this
+using loaded in data arrays. However, given the size of the individual raster files,
+this module provides the `process()` method that calculate statistics for
+a given area. For this, we need to provide a geodataframe and column to aggregate to.
+
+Let's load the second-level administrative boundaries for Ethiopia from our country
```
Tinkaa, line 111: Ha I am such a detail-nerd, but this states second-level and you code uses zero-level

docs/source/datasources/usgs_ndvi.rst
```diff
@@ -0,0 +1,131 @@
...
+        gdf=eth_gdf0,
+        feature_col="ADM0_EN"
+    )
+
+We can then load in and use the calculated statistics using `load()`. Since the
+processed file is saved based on the `feature_col` name, we have to pass
+`feature_col` to load in the correct data.
+
+.. code-block:: python
+    ndvi_smooth.load(feature_col="ADM0_EN")
```
Tinkaa, line 131: nice docs!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -536,6 +516,22 @@ def _download(self, filepath: Path, url_filename: str, clobber: bool):
...
+                "%a, %d %b %Y %X %Z",
+            )
+            if not clobber and url_time <= file_time:
+                logger.info(
+                    f"File {filepath} exists, clobber set to False, and "
+                    "file not been modified since last download. "
+                    "Using existing files."
+                )
+                return filepath
+
```
Tinkaa, line 534: love it!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -619,32 +620,40 @@ def _determine_process_dates(
...
             else:
                 # remove processed dates from dates to process
                 dates_to_process = [
                     d
                     for d in dates_to_process
-                    if d not in dates_already_processed
+                    if d not in list(dates_already_processed.keys())
+                    or self._get_modified_time(year=d[0], dekad=d[1])
+                    > dates_already_processed[d]
+                ]
```
Tinkaa, line 648: also love it!

Tinkaa (approved):
Looking great!

docs/source/datasources/usgs_ndvi.rst
caldwellst: God yeah, sorry, changed it because even second level was taking time!

docs/source/datasources/usgs_ndvi.rst
caldwellst: Ah no it isn't.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Had [this issue](https://stackoverflow.com/questions/56636172/why-is-git-rebase-flagging-conflicts-on-new-branch-specific-files) with `git rebase develop` flagging new files repeatedly for conflicts. The SO post helped fix it.

caldwellst: No longer using `check_file_existence` anyway, though, as we are doing much more complex checks based on modification time.

caldwellst:
All comments resolved! Just testing left.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
+"""Class to download and process USGS FEWS NET NDVI data.
+
+Data is downloaded from the `USGS FEWS NET data portal
```
turnerm, line 3: This is all duplicated in `docs/` correct? I would remove it from here then, otherwise it will be impossible to maintain them both. Or just keep the bare minimum here.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+
+import aatoolbox.utils.raster  # noqa: F401
+from aatoolbox.config.countryconfig import CountryConfig
+from aatoolbox.datasources.datasource import DataSource
+from aatoolbox.utils.dates import (
+    _compare_dekads_gt,
+    _compare_dekads_lt,
+    _dekad_to_date,
+    _expand_dekads,
+    _get_dekadal_date,
```
turnerm, line 62: Sorry, I can't find the conversation I saw you had with Tinka on this, but if I recall correctly there was a link to an SO post, and one of the answers basically said that if you want modules to not be part of the API, just don't put them in `__init__`.  

The leading underscore for a function denotes that it's internal within its own module. So if you're going to import these functions in other modules, the leading underscores should be removed.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+        super().__init__(
+            country_config=country_config,
+            datasource_base_dir="usgs_ndvi",
+            is_public=True,
+            is_global_raw=True,
+            config_datasource_name="usgs_ndvi",
+        )
+
+        # set area url and prefix from config
+        if self._country_config.usgs_ndvi is None:
```
turnerm, line 117: This shouldn't be needed as it's taken care of when you pass `config_datasource_name` to the parent

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+            raise AttributeError(
+                "The country configuration file does not contain "
+                "any USGS NDVI area name. Please update the config file and "
+                "try again. See the documentation for the valid area names."
+            )
+        self._area_url = self._datasource_config.area_url
+        self._area_prefix = self._datasource_config.area_prefix
```
turnerm, line 124: These should not be needed anymore (in case you have them for MyPy)

src/aatoolbox/datasources/datasource.py
```diff
@@ -67,7 +67,7 @@ def _config_attribute_name_validator(
             return getattr(country_config, config_datasource_name)
```
turnerm, line 67: Oops while you're here, can you make this a @staticmethod, thanks so much (my bad)

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+            file_year = f"{year:04}_"
+        else:
+            file_year = f"{year-2000:02}"
+        file_name = (
+            f"{self._area_prefix}{file_year}"
+            f"{dekad:02}{self._data_variable_suffix}"
+        )
+        return file_name
+
+    def _load(self, filepath: Path, drop_modified: bool = False):
```
turnerm, line 710: @staticmethod

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+
+        # get dates for processing
+        all_dates_to_process = _expand_dekads(
+            y1=self._start_year,
+            d1=self._start_dekad,
+            y2=self._end_year,
+            d2=self._end_dekad,
+        )
+
+        if stats_list is not None:
```
turnerm, line 303: I think it would be simple to create a combined stats / percentile list here:
`full_stats_list = stats_list + [f"{p}quant" for p in percentile_list]`
Then you can simply loop through this, and pass it to a single parameter called `stat` in `_process`

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+        if stats_list:
+            stat = stats_list[0]
+        elif percentile_list:
+            stat = f"{percentile_list[0]}quant"
+        processed_path = self._get_processed_path(
+            feature_col=feature_col, stat=stat
+        )
+
+        if processed_path.is_file():
+            dates_to_process, df = self._determine_process_dates(
```
turnerm, line 568: maybe call this `df_already_processed` or something like that

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,851 @@
...
+                stats_list=stats_list,
+                percentile_list=percentile_list,
+                **kwargs,
+            )
+            data.append(stats)
+
+        # join data together and sort
+        df = pd.concat(data)
+        df.sort_values(by="date", inplace=True)
+        df.reset_index(inplace=True, drop=True)
```
turnerm, line 605: You can chain these bad boys if you take out `inplace=True`. I thought you would never miss such an opportunity!

requirements/requirements-dev.txt
```diff
@@ -30,9 +30,7 @@ basicauth==0.4.1
     #   -r requirements/requirements.txt
```
turnerm, line 30: I think the reqs files shouldn't have any changes

turnerm (approved):
Looks incredible! I just left a few final minor comments. 

I do have one question though, which is, what happens in the following scenario:
- Download data which happens to be unsmoothed
- Process it 
- A couple of days later, download the data again, this time it is smoothed so the raw data is overwritten
- Run `.process()`
Would the previous processed data get overwritten, as it should?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: That's what I thought, but it was still in `fewsnet.py` so was wondering maybe it is required for some reason.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Then I guess in `_process()` we reprocess the stat passed in to see if it's a percentile or a stat, so which way to pass to the `compute_raster_stats()`. Maybe this is a longer term issue that we sort out by changing the parameters passed to `compute_raster_stats()`? As it stands, you have to separately pass the stats or percentile.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: My god. Of course.

requirements/requirements-dev.txt
caldwellst: Am having the same problem as @castledan in #116, where running `pre-commit run --all-files` changes requirements. It is also based on local environment, tested with/without `venv` and get different results.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Yeah, agreed. In that case, should we make the module "private" then with `_dates.py`? That was based on [this post](https://stackoverflow.com/questions/3602110/python-private-module-in-a-package).

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Is there a way to reference the additional documentation that is elsewhere?

requirements/requirements-dev.txt
turnerm: Yes sorry this is totally my bad, you're right that it depends on your Python version, are you using 3.9? I'm using 3.8 but should probably upgrade, so you can keep these changes and we'll make sure it's explicit in Daniele's PR

caldwellst: Yeah, I'm using 3.9.

caldwellst: Thanks for clarifying, was about to manually remove but will leave as is!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
turnerm: Ah I'm so sorry, my bad again, I thought that `_process` didn't differentiate between a normal stat and percentile after appending `quant`, but indeed it's important for passing to `compute_raster_stats`. 

A cleaner way to do this might be to just have an arg for a single stat (as a string), and another arg for a boolean that specifies `is_percentile`. But it's also fine the way it is.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
turnerm: Oops that should be deleted, maybe you can sneak it in here?

caldwellst:
With regards to your hypothetical @turnerm, in your scenario, we would reprocess the data (woop!). Basically, I adjusted the `load_raster()` method to add the modification time as a dimension to the raster file, so it's transparent to user if loading when the file was updated and also, it is automatically brought into the processed data frame by default by `compute_raster_stats()` (it just becomes a column called `modified`).

Then, in `_determine_process_dates()`, if `clobber`, then we just re-process all files. If not `clobber`, then we compare the file modification time to the modification time stored in the processed file, and re-process if the raw file has been updated more recently.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Yeah can do!

caldwellst: Would you mind pointing me to the code where this validation is done? Interested to see how you made the setup work, it's cool code but not entirely clear on how it's working ha!

caldwellst: I guess it's done in the `_config_attribute_name_validator`.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: Love the idea, have changed!

turnerm:
@caldwellst ah sorry I had not picked that up, sounds great!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
turnerm: You could just link directly to the read the docs page. Maybe we should start doing that for all the data sources.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
turnerm: Yeah exactly!

caldwellst: This isn't actually raising an `AttributeError` in the FEWS NET test because `getattr()` doesn't return an error when it's `None`, the default.

caldwellst:
Okay @Tinkaa @turnerm I hope this is close to gold now!! I won't be able to get to testing this week, but can get to it when I'm back. Let me know if we should leave this here until testing is able to be completed or if that should be done in a separate PR. Will continue to address any more comment or suggestions you have other than testing, though, so it's ready to go as soon as testing implemented (presuming that doesn't identify a myriad of errors).

src/aatoolbox/datasources/usgs/usgs_ndvi.py
Tinkaa: Awesome, thanks for taking care of this for FN as well!

Tinkaa:
I think it is [caterpillar mushroom](https://dsfantiquejewelry.com/blogs/journal/top-10-unknown-things-that-are-much-more-valuable-than-gold), great job! I would def do the testing in a separate PR but maybe one that builds on top of this one?

caldwellst:
What the hell is caterpillar mushroom hahahahaha!?

src/aatoolbox/datasources/usgs/usgs_ndvi.py
turnerm: No, neither the module or the functions should be private, as they are used in other modules. But they are "hidden" from the user by not being in the top level __init__.

turnerm:
I can't find the comment now but good find with the FewsNet issue @caldwellst ! I've addressed the issue in #118

Tinkaa:
>What the hell is caterpillar mushroom hahahahaha!?

It is worth two times more than gold, i.e. your work is even better than gold. Ahah okay back to the serious stuff.

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,815 @@
...
+            feature_col=feature_col, stat=stat
+        )
+
+    def _find_processed_files(self, feature_col: str) -> Generator:
+        base_file_name = self._get_processed_base_filename(
+            feature_col=feature_col
+        )
+        return self._processed_base_dir.glob(f"{base_file_name}_*.csv")
+
+    def _get_url(self, filename) -> str:
```
turnerm, line 776: I feel like this might be an important method to test -- if instead of mocking out `_USGSNdvi._download`, you mock out `urlopen` and `ZipFile` then it should be pretty straightforward I think....

tests/datasources/test_ndvi.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def test_process_and_load(
+    mocker, mock_ndvi, mock_aa_data_dir, mock_country_config, gdf, da
+):
+    """Test processing NDVI values."""
+    ndvi = mock_ndvi(variable="smoothed")
+
+    # TODO: now created `load_raw` to be able to mock but would like
+    #  to do it from xr.load_dataset directly
```
turnerm, line 160: You can mock it directly, the reason why it didn't work for Tinka in the IRI tests is because `xr.load_dataset` is also needed in the testing module itself, but ends up being mocked too

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,815 @@
...
+        df = df.loc[keep_rows]
+
+        # put feature column as 1st column
+        df_feature_col = df.pop(feature_col)
+        df.insert(loc=0, column=feature_col, value=df_feature_col)
+
+        return df
+
+    def load_raster(
+        self, date: Union[date, str, Tuple[int, int]]
```
turnerm, line 362: Oops should have caught this earlier but you should rename the variable `date` since you also import `date` from `datetime`, to avoid confusion

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,815 @@
...
+        # join data together and sort
+        df = pd.concat(data).sort_values(by="date").reset_index(drop=True)
+
+        # saving file
+        self._processed_base_dir.mkdir(parents=True, exist_ok=True)
+        df.to_csv(processed_path, index=False)
+
+        return processed_path
+
+    def _determine_process_dates(
```
turnerm, line 576: Would this be hard to test, it's a lot of logic

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,815 @@
...
+        FileNotFoundError
+            If the requested file cannot be found.
+        """
+        year, dekad = get_dekadal_date(input_date=date)
+
+        filepath = self._get_raw_path(year=year, dekad=dekad, local=True)
+        try:
+            da = rioxarray.open_rasterio(filepath)
+            # get time file was updated
+            file_time = datetime.fromtimestamp(filepath.stat().st_mtime)
```
turnerm, line 390: This block should be moved out of the try statement

turnerm (commented):
Hooray for tests! Just a couple of small comments plus asking for a couple of more tests (sorry) but if they're too complicated then don't worry too much about it

src/aatoolbox/datasources/usgs/usgs_ndvi.py
```diff
@@ -0,0 +1,815 @@
+"""Class to download and process USGS NDVI data.
```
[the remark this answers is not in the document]
caldwellst: Not nitpicking, clear and good idea!

src/aatoolbox/datasources/usgs/usgs_ndvi.py
caldwellst: I think we can just test this method on its own, because we can't validate whether the URL is correct, so just validating the attributes and filename are passed correctly.

tests/datasources/test_ndvi.py
caldwellst: Aha I even figured it out relatively quickly. Dream life.

turnerm (approved):
Thanks for adding the extra tests, it looks good to merge! My only comment for next time is that you usually want to avoid testing private methods, instead try to write the test for the public method that accesses it, that way you ensure testing of the interface.

Issue #105, opened 2022-04-18
Adjust FEWS NET based on custom config PR


Tinkaa:
After the custom config PR has been merged, two things in the fewsnet script have to be adjusted:
- use the `_dataset_config`
- create tests for the `countryconfig.py` part that is specific to fewsnet.

turnerm:
Closed by #107

PR #107, opened 2022-04-19
add tests config and datasource validator

Tinkaa:
Adresses #105 

Also added a `test_datasource.py`. Let me know if you think it is overkill or should actually have more tests.


turnerm:
Hey @Tinkaa, sorry, but I think @caldwellst is about to merge the [method decorator PR](https://github.com/OCHA-DAP/pa-aa-toolbox/pull/99). Do you think you could update the FewsNet download function to be a class method?

turnerm (approved):
Looks great so far!

Tinkaa:
@turnerm good one, now updated download function to be a class method

PR #111, opened 2022-04-21
Update requirements and support 3.10

turnerm:
Quick PR before the release:
- update the requirements (should probably find a better way to keep them fresh)
- support Python 3.10 by making a rasterio requirement to use the latest beta


Tinkaa (approved):
Looks goodio

PR #113, opened 2022-04-22
Fix FEWS NET clobber

turnerm:
Great catch @caldwellst that clobber was not implemented properly. I've tried to fix it and also add a test.

I'm wondering if we should always be testing our clobber implementations... seems like a  huge pain though.

The changes are `raster.py` are because mypy suddenly started complaining!

### DIFF
--- a/src/aatoolbox/datasources/fewsnet/fewsnet.py
+++ b/src/aatoolbox/datasources/fewsnet/fewsnet.py
@@ -128,20 +128,20 @@ class FewsNet(DataSource):
         >>> eth_fn_202106_path = fewsnet.download(pub_year=2021,pub_month=6)
         """
         self._check_date_validity(pub_year=pub_year, pub_month=pub_month)
-        pub_month_str = f"{pub_month:02d}"
+        pub_month_str = self._get_pub_month_str(pub_month)
         # we prefer the country data as this more nicely structured
         # thus first check if that is available
         try:
             return self._download_country(
                 pub_year=pub_year,
-                pub_month=pub_month_str,
+                pub_month_str=pub_month_str,
                 clobber=clobber,
             )
         except zipfile.BadZipFile:
             try:
                 return self._download_region(
                     pub_year=pub_year,
-                    pub_month=pub_month_str,
+                    pub_month_str=pub_month_str,
                     clobber=clobber,
                 )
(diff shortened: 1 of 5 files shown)


src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -371,35 +380,37 @@ def _download_zip(
...
         )
         if country_dir.is_dir():
             return country_dir
         elif region_dir.is_dir():
             return region_dir
 
         raise FileNotFoundError(
-            f"No data found for {pub_year}-{pub_month} covering "
+            f"No data found for {pub_year}-{pub_month_str} covering "
             "{self._country_config.iso3} "
```
caldwellst, line 405: This and one below should be f-strings.

src/aatoolbox/datasources/fewsnet/fewsnet.py
```diff
@@ -209,17 +209,17 @@ def load(  # type: ignore
...
             dir_path=dir_path, projection_period=projection_period
         )
         return gpd.read_file(file_path)
 
     @staticmethod
-    def _check_date_validity(pub_year, pub_month):
+    def _check_date_validity(pub_year: int, pub_month: int):
         try:
             pub_date = datetime.datetime(year=pub_year, month=pub_month, day=1)
         except ValueError as err:
```
caldwellst, line 225: Line 227 below this should be an f-string.

caldwellst: Also line 233

caldwellst (approved):
All looks good! Was wondering about the testing. I was wondering, since the `@check_file_existence` decorator is always used on internal functions/methods, at least so far, could we require that `clobber` is passed, and raise a key error if not, so that if functions don't use clobber then in the internal method they can explicitly pass `True` or `False` even if `clobber` isn't an argument in the parent function?

src/aatoolbox/datasources/fewsnet/fewsnet.py
turnerm: Fixed!

turnerm:
@caldwellst great idea to make clobber required!

caldwellst (approved):

PR #115, opened 2022-04-22
Bugfix: upgrade black version in precommit

castledan:


turnerm (changes requested):
The pre-commit file change looks good, but the requirements files shouldn't be changing at all.

PR #116, opened 2022-04-22
Bugfix: upgrade black version in precommit

castledan:
@turnerm I created a new pull request, with only the changes in the pre-commit file. However, if you run `pre-commit ...` in this branch (and in develop), also the requirements files will change.


turnerm (approved):
@castledan actually the requirements problem is totally my bad, I realized that they change depending on the Python version that you use. I'm guessing you're on 3.9? I should definitely upgrade in that case. Let me make a separate PR that makes this issue more explicit.

PR #117, opened 2022-04-25
Update requirements to Python 3.9

turnerm:
@castledan discovered that the requirements files depend on the Python version used. I've since upgraded from 3.8 to 3.9, and made the following changes in this PR:
- upgraded the reqs files
- added the header to these files, so it's clear which Python version is used
- stated in CONTRIBUTING.md that we are using Python 3.9 for development
- updated the pip-tools version, due to the same issue that Daniele is solving in #116


caldwellst (commented):
So, no direct comments on the code, however, running `pre-commit run --all-files` is still generating differences on my machine for all the requirements files. It primarily seems to center around `dask`, where my machine is removing `distributed` and `bokeh` as requirements, and a cascading list of dependencies from there also gets removed.

caldwellst:
[git_diff.txt](https://github.com/OCHA-DAP/pa-aa-toolbox/files/8552326/git_diff.txt)

This is my `git diff`.

turnerm:
@caldwellst Oh man this is making me crazy :zany_face: 

I'm trying to trace the requirements:
- xarray [requires dask complete](https://github.com/pydata/xarray/blob/main/setup.cfg#L103)
- dask complete [requires bokeh, distributed, jinja2](https://github.com/dask/dask/blob/main/setup.py#L14-L25)

I think most of the differences in your diff come from the dask complete requirements. 

But are you using Python 3.9 or 3.10? Because I noticed your rasterio is the version required for 3.10.

caldwellst:
So, wanted to double check everything is set up correctly, used [this post](https://github.com/pre-commit/pre-commit/issues/1504) just to guide myself on how to check. But yeah, seems `pre-commit` is using 3.9.7

```shell
$(head -1 $(which pre-commit) | cut '-d!' -f2) --version
Python 3.9.7
```

So I have tested on a new `venv` with 3.9.0. And when I did that!

```shell
$(head -1 $(which pre-commit) | cut '-d!' -f2)
Python 3.10.2
```

So, clearly it's something wrong with my system where I've not got `pre-commit` matching my `venv`. I don't have time to look deeper now, but will solve this for myself in the future by digging deeper (or if y'all know how to solve, advice appreciated!).

Issue #110, opened 2022-04-21
Testing for data source in config


turnerm:
@Tinkaa I realized that for [this test](https://github.com/OCHA-DAP/pa-aa-toolbox/blob/develop/tests/datasources/test_datasource.py#L41) in `DataSource`, we should be testing through the child classes instead

turnerm:
I was wrong, closed with #118

PR #118, opened 2022-04-25
Fix datasource config attribute bug

turnerm:
@caldwellst found a bug where the data source config checker was not actually working. This is because if an attribute is a default option, it is actually set to `None` and an attribute error won't be raised. The changes here are:
- have the method check for both an attribute error and `None`
- remove the explicit check in FewsNet
- add a check in CodAb
- remove the DataSource test as it should be tested in the child classes 

Addresses #110


caldwellst (approved):
All looks good. I'm just wondering, what's the logic behind testing for each child class rather than just having a single test?

src/aatoolbox/datasources/datasource.py
```diff
@@ -60,16 +60,23 @@ def __init__(
...
-        except AttributeError as err:
+            datasource_config = getattr(country_config, config_datasource_name)
+        except AttributeError:
+            datasource_config = None
+        # If the datasource is one of the defaults, it's set to None and
+        # thus an attribute error won't be raised. So also need to check for
+        # the case when it is None.
+        if datasource_config is None:
             raise AttributeError(
                 f"{config_datasource_name} needs to be added to the the "
```
Tinkaa, line 76: very minor, but double "the" in error string

Tinkaa (commented):
Thanks for taking care of this! 

Same question as Seth. Seems like duplication to have a test for it for each datasource? 

If not, we should also add the test to `test_fewsnet.py`.

src/aatoolbox/datasources/datasource.py
turnerm: Oh man I do this this all the time, good catch

turnerm:
> what's the logic behind testing for each child class rather than just having a single test?
> Same question as Seth. Seems like duplication to have a test for it for each datasource?

Yes good question. I think in general it's considered best practice to test from the public API as much as possible (there are many good discussions of this online, see e.g. [here](https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice)). 

In this case I would consider `DataSource` as private as it's not something we generally intend to expose. But I think you guys are right actually that testing in more than one child class is not necessary.

And in fact, writing the above, I realized that maybe we _do_ want to make `DataSource` (unoficially) public, since e.g. it's used for extension. So maybe I will add the tests back there instead. What do you think?

> If not, we should also add the test to test_fewsnet.py.

It's already added.

Tinkaa (approved):
Hmm okay now you are making me doubt again. That argumentation does make sense (becoming such design gurus). But I would agree on your point of having `DataSource` semi-public so then it makes sense to test it there? I would say this looks good

caldwellst:
In my head it makes sense to keep it just on `DataSource`, but you guys are becoming design experts so will leave the final decision to y'all!

PR #121, opened 2022-04-28
add language_version to pre-commit pip-compile

Tinkaa:
While trying to switch to python 3.9, I was having the same issue as Seth had. Namely that my version of pre-commit wasn't the same as my python version in my venv. 

Now it seems, you can force pre-commit to use a certain version, see [here](https://pre-commit.com/#overriding-language-version). This solved the issue for me and to me this seems like a logical thing to do. 

Let me know what you think!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,20 +46,28 @@ repos:
     hooks:
       - id: pip-compile
         name: pip-compile requirements.txt
+        language: python
+        language_version: python3.9
         files: setup.cfg
         args: [setup.cfg, -q, -o, requirements/requirements.txt]
       - id: pip-compile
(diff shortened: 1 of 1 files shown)


turnerm (approved):
Amazing find @Tinkaa! It looks great, but what do you think about just setting `default_language_version` for all the hooks?

Also can you add a pre-commit [badge](https://pre-commit.com/#badging-your-repository) :D

Tinkaa:
Oh wow that is so much better. I shouldn't have been so lazy and read a bit more. Thanks! Changed now.

castledan:
Thank Tinka, and I agree with Monica's solution!

castledan (approved):

caldwellst:
Amazing Tinka!

PR #122, opened 2022-04-29
update hdx version

Tinkaa:
While trying to download a hdx data source (in the `pa-global-monitoring`), I realized there is a bug in hdx-python-api 5.6.2, which has been fixed in 5.6.4. So we should force at least this version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+default_language_version:
+    python: python3.9
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.0.1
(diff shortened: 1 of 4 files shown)


turnerm (commented):
Ah shoot, can I ask, what's the bug? 

Since it's affecting the stable version, we should probably make it into a hotfix and merge it into main. In that case, you will also need to update the Changelog and bump the version to 0.4.1. Then once it's merged, make a release, I can show you how to do that.

Tinkaa:
> Ah shoot, can I ask, what's the bug?

`TypeError: download_file() takes 2 positional arguments but 4 were given`
Simple example below
``` 
from aatoolbox.utils.hdx_api import load_dataset_from_hdx
load_dataset_from_hdx(
    hdx_address="cod-ab-uga",
    hdx_dataset_name="uga_admbnda_ubos_20200824_SHP.zip",
    output_filepath="bla")
```

I am pretty sure that it is because in the hdx api 5.6.2 downloader some args are not given as kwargs. See the diff between 5.6.2 and 5.6.4 [here](https://github.com/OCHA-DAP/hdx-python-api/compare/5.6.2...5.6.4)

Do you also have this error on 5.6.2?

Tinkaa:
Closing this as this PR is replaced by #126

Issue #123, opened 2022-05-09
Loading zipped shapefiles on Windows


turnerm:
@PaulineNimo found the following bug on Windows when running `codab.load()`:

```
Traceback (most recent call last):
  File "fiona\_shim.pyx", line 83, in fiona._shim.gdal_open_vector
  File "fiona\_err.pyx", line 291, in fiona._err.exc_wrap_pointer
fiona._err.CPLE_OpenFailedError: '/vsizip//G:\Shared drives\Predictive Analytics\CERF Anticipatory Action\General - All AA projects\Data\public\raw\npl\cod_ab\npl_cod_ab.shp.zip\npl_admbnda_adm1_nd_20201117.shp' does not exist in the file system, and is not recognized as a supported dataset name.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\pauni\Desktop\Work\OCHA\Nepal\analysis\nplanalysis\lib\site-packages\aatoolbox\datasources\codab\codab.py", line 162, in _load_admin_layer
    return gpd.read_file(f"zip:///{self._raw_filepath / layer_name}")
  File "C:\Users\pauni\Desktop\Work\OCHA\Nepal\analysis\nplanalysis\lib\site-packages\geopandas\io\file.py", line 201, in _read_file
    with reader(path_or_bytes, **kwargs) as features:
  File "C:\Users\pauni\Desktop\Work\OCHA\Nepal\analysis\nplanalysis\lib\site-packages\fiona\env.py", line 408, in wrapper
    return f(*args, **kwargs)
  File "C:\Users\pauni\Desktop\Work\OCHA\Nepal\analysis\nplanalysis\lib\site-packages\fiona\__init__.py", line 267, in open
    c = Collection(path, mode, driver=driver, encoding=encoding,
  File "C:\Users\pauni\Desktop\Work\OCHA\Nepal\analysis\nplanalysis\lib\site-packages\fiona\collection.py", line 162, in __init__
    self.session.start(self, **kwargs)
  File "fiona\ogrext.pyx", line 540, in fiona.ogrext.Session.start
  File "fiona\_shim.pyx", line 90, in fiona._shim.gdal_open_vector
[...]

PR #124, opened 2022-05-10
Fix COD AB loading in Windows

turnerm:
Solves #123. Need confirmation from @PaulineNimo that it works before merging


Tinkaa (approved):

PaulineNimo:
> Solves #123. Need confirmation from @PaulineNimo that it works before merging

I can confirm that it is working well and was able to read the shapefile. Thank you, Monica.

PaulineNimo (approved):
Great job Monica!

PR #125, opened 2022-05-12
update eth adm bound name

Tinkaa:
The ethiopia config file was outdated, so updated it. 

One problem: the admin0 filename follows a different structure than the others.. I now added it as a custom name, but it can be confusing to users. I was thinking that we should maybe also require a `min_level` but it might overcomplicate things as well. Do you have thoughts?


turnerm:
@Tinkaa so sorry for dragging my heels on this, I think I was avoiding it because I was dreading the moment that COD AB file edge cases would come to terrorize us. This is such an annoying one! 

Okay don't hate me for this suggestion, but what I would do is make an optional parameter for admin0. Then, if admin0 is not provided explicitly, use the layer base name, otherwise use the admin0 name.

The question is whether to put the logic for this in the config file or the CodAB class -- I'm thinking the class?

Tinkaa:
Hello! Finally looking at this again. 

My intuition would be to keep it in the config file, since I think it is gonna be more confusing if you have two places where you can indicate codab filenames. What would be your reason to put it in the CodAB class? 

Secondly, I had another idea for how to manage this, looking at the wider picture. Because we now have a parameter named `admin_level_max` but would it be an idea to instead have a parameter named smth like `admin_level_available` which is then a list of the available admin levels that follow the general filename pattern
Then all admin levels that don't follow this pattern, need to provide a custom filename. What do you think?

turnerm:
> My intuition would be to keep it in the config file, since I think it is gonna be more confusing if you have two places where you can indicate codab filenames. What would be your reason to put it in the CodAB class?

I honestly don't remember, aha. Maybe not to clutter the config class too much. But perhaps you are right! 

> Secondly, I had another idea for how to manage this, looking at the wider picture. Because we now have a parameter named admin_level_max but would it be an idea to instead have a parameter named smth like admin_level_available which is then a list of the available admin levels that follow the general filename pattern
Then all admin levels that don't follow this pattern, need to provide a custom filename. What do you think?

I love this so much!

Tinkaa:
@turnerm ready for review

turnerm:
Hey Tinka, this looks really great! My only suggestion would be (which we can also just turn into an issue) to create an optional `admin0_name` parameter in `CodABConfig`. Then what would happen is if someone requests admin 0, first `layer_base_admin_levels` is checked, if it's not there, check for `admin0_name`, if not there then throw an error.

The `custom_layer_names` parameter was created for levels that don't fit into the numbered paradigm, e.g. Nepal has some unique district levels. Let me know if that makes sense. But again, definitely fine to make an issue for the future as I know you're tight on time. Otherwise looks good to merge!

turnerm (approved):

Tinkaa:
Thanks for having a look! 
Hmm I am not 100% sure if I agree.. I get what you mean, but then there is gonna be so many exceptions. Like what if level 0,1, and 3 follow the layer_base_name but level 2 doesn't. Therefore my intuition would be to just have on `layer_base_name` structure and all layers that don't follow that format, go into `custom_layer_names`. How would you envision this situation, should we also have an `admin2_name` etc?

turnerm:
Replying now to the ghost of Tinka -- basically my idea is to `custom_layer_names` to anything that isn't actually an admin level (such as the Nepal districts, not sure if there are any other examples!). Then admin 0 to 4 are set directly in the config file, either using the `layer_base_name`, or manually by the user if they are exceptional. 

I also kept `admin_level_max` for now, I don't think it's great but I prefer it to specifying a list, which I think can be more error-prone. 

See my implementation in #162.

PR #134, opened 2022-10-14
Drop Python 3.6 support

### LINKED ISSUE #132: Drop Python 3.6 support
I didn't realize Python 3.6 reached EOL last year, maybe we can make our own lives easier by dropping it

turnerm:
Resolves #132. 

I also:
- took out the `rasterio` req for 3.10 as it's not needed anymore
- (I think) fixed a bug in io.py where it would not have worked for 3.7

### DIFF
--- a/.github/workflows/run-python-tests.yaml
+++ b/.github/workflows/run-python-tests.yaml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


caldwellst (approved):
Oh nice catch on the `io.py`. All looks good to me so I think we can go ahead with this, then I'll merge into the NDVI branch so you can review.

Issue #129, opened 2022-05-19
Generalize date object / string argument


turnerm:
From discussion in #128, want to make it easy for data sources to have `start_date` and `end_date` parameters that accept both a date object and ISO 8601 string

caldwellst:
Addressed by #138

Issue #128, opened 2022-05-17
Standardizing input dates


turnerm:
I couldn't find the discussion on Slack, but iirc we wanted to talk about the standardization of input dates, seeing as most data sources require one type of date or another as input. 

For the sake of this discussion (and simplicity), I would say try to stick to our own use cases - we can always add / modify if others come up. I think our main use case involves downloading historical data over several years, often the full range of what's available. 

Thus, I will propose a couple of options for `start_date` and `end_date` parameters (with the default set to the range of what's available for the data source):

1. Passing in just the year as an int or str. The advantage is that it's very simple. However, the downside is that you lose any date / month flexibility, and it seems a bit rigid as a paradigm for all (future) data sources. 

2. A Python `date` object. While it's a bit annoying to create such an object, we are probably not using the parameter that often anyway. And it does give some flexibility if needed. 

FEWS NET is a special case -- if we use option 1 then it can stay the same, with option 2 maybe it can just take the year / month and ignore the day number of the `date` object. @Tinkaa what are your thoughts?

If we go with 2, we could also support ISO 8601 format as @caldwellst did for NDVI, but we should perhaps write some utils for that if we're going to use it across data sources.

caldwellst:
Hey @turnerm, thanks for starting the discussion! Sorry, my time for toolbox got eaten up last week and haven't had time to return to it yet.

My initial thought would be to go with the second option as a *baseline*. So all implementations have input dates based on `start_date` and `end_date`, where ISO8601 strings and `date` objects are passed in. I put in a `_dates` module in the NDVI PR that could easily have functions to deal with this (split off from the current dekadal functionality). However, as you mention, there are special cases for case 2. NDVI just needs years/dekads, so it requires conversion to that format. FEWSNET from @Tinkaa and monthly CHIRPS from @castledan require years/months. So I think allowing those additional inputs as tuples would make sense.

I think the flexibility makes sense, because at least in the case of dekads, it's much easier to just say the 1st of a month or some other date rather than calculating the dekad yourself. And my doing the conversion on the backend through the toolbox it again allows a single date to be passed to all methods, even if each of them might allow a specific tuple format to also be passed in. Let me know what y'all think!

caldwellst:
Just to add for option 1, because left unsaid up above, for data like NDVI or other large rasters, enforcing year downloads and processing can take a lot of time, so probably best to avoid that if we can.

castledan:
Can I suggest a third option?
[...]

PR #138, opened 2022-11-10
Implement string date function

caldwellst:
Reviewing your GLOFAS module inspired me to tackle some of the backlog in particular #128 and #129.

### DIFF
--- a/src/aatoolbox/datasources/usgs/ndvi_base.py
+++ b/src/aatoolbox/datasources/usgs/ndvi_base.py
@@ -461,10 +461,11 @@ class _UsgsNdvi(DataSource):
         local_filename = filepath.stem
 
         url = self._get_url(filename=url_filename)
+        year, dekad = self._fp_year_dekad(filepath)
+
         try:
             resp = urlopen(url)
         except HTTPError:
-            year, dekad = self._fp_year_dekad(filepath)
             logger.error(
                 f"No NDVI data available for "
                 f"dekad {dekad} of {year}, skipping."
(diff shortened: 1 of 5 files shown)


turnerm (commented):
This is so great Seth, thanks for implementing it! Just one small bug to fix. 

It _did_ make me notice a couple of additional things that would also be very quick fixes and great to throw in, if you wouldn't mind (sorry):

- For NDVI, there are no logger messages for downloading / processing, would you be able to add them? (e.g., {year} and {dekad} is being downloaded to {filepath}). If you're worried about logger spam you can make it `DEBUG`. 
- In the `_Chirps` class I noticed that there are a lot of start dates / end dates being passed around where I think the instance variables can just be used -- if I'm not wrong, would you mind changing the methods to use instance variables? Thanks so much. (Unless @castledan there is a reason why you didn't do it this way?)

caldwellst:
Thanks Monica! Have added in the logger messages to NDVI and used instance variables across CHIRPS (all tests still passing and reviewing the code I saw no reason not to), as well as addressed the other comments!

castledan:
Thank you Seth for fixing the CHIRPS code. I'm not sure why I'd implemented it in a different way, I don't think there was a particular reason.

src/aatoolbox/datasources/usgs/ndvi_base.py
```diff
@@ -487,6 +488,11 @@ def _download(self, filepath: Path, url_filename: str, clobber: bool):
                 )
                 return filepath
 
+        logger.info(
+            f"Downloading NDVI data for {year}, dekad {dekad} "
+            "into {filepath}."
```
turnerm, line 493: Missing the `f` here

turnerm (approved):
Looks good to me, go ahead and merge (just one typo to fix). I will merge this into my own GloFAS branch.

Issue #130, opened 2022-05-30
Raster in top-level __init__


turnerm:
We should make the raster mixin importable from the top level and add to the readme, since we use it in a lot of pilot work

caldwellst:
Addressed in #139

Issue #42, opened 2021-12-14
Distinguishing between dataset names / authors and providers


turnerm:
At some point @Tinkaa and I realized that we need to make the distinction between dataset names / authors and providers. Some examples:
- The code for downloading the COD ABs is under `datasources/codab`, but the data is downloaded from HDX, using code from `utils/hdx_api`
- The GloFAS data comes from CDS, and we currently have `datasources/cds` 
- ECMWF data has multiple providers: CDS, their API, and AWS

I would propose to have a `dataproviders` package at the same level as `datasource`, and place all data provider functionality there.

Tinkaa:
Oeff I am re-seeing this issue and I think it something we should discuss soonish. As we now e.g. also have `iri` and `usgs`. We use them as top-level dirs in `datasources`, but we could also choose to use `ndvi` instead of `usgs` for example. I don't know which one is better, but I think we have to be consistent.

caldwellst:
Important discussion to have. Monica brought it up again in #139. I think too much for that specific PR but we should consider moving forward. I think it will be clearer when we add WRSI which would also be (at least easiest) using USGS data.

PR #139, opened 2022-11-10
Document and add raster to main init, address #130

caldwellst:
Addressing #130. Wanted to get more familiar with sphinx and our documentation. I'm sure it isn't perfect but hope there isn't too much changes necessary and this is useful!

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -1,6 +1,14 @@
 AA Toolbox API
 ==============
 
+Configuration
+-------------
+
+.. automodule:: aatoolbox.config.countryconfig
+
+.. autofunction:: aatoolbox.create_country_config
+.. autofunction:: aatoolbox.create_custom_country_config
+
 Data sources
 ------------
 
@@ -8,77 +16,104 @@ CHIRPS
 ^^^^^^
 
 .. automodule:: aatoolbox.datasources.chirps.chirps
-   :members:
-   :undoc-members:
-   :show-inheritance:
+
+Daily
+"""""
+
+.. autoclass:: aatoolbox.ChirpsDaily
+   :inherited-members:
+
+Monthly
+"""""""
+
+.. autoclass:: aatoolbox.ChirpsMonthly
+   :inherited-members:
 
 Common Operational Datasets
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. automodule:: aatoolbox.datasources.codab.codab
+.. autoclass:: aatoolbox.CodAB
    :members:
-   :undoc-members:
-   :show-inheritance:
 
 FEWS NET
 ^^^^^^^^
 
 .. automodule:: aatoolbox.datasources.fewsnet.fewsnet
+.. autoclass:: aatoolbox.FewsNet
    :members:
-   :undoc-members:
-   :show-inheritance:
 
 IRI
 ^^^
 
-Seasonal forecasts
-""""""""""""""""""
-
 .. automodule:: aatoolbox.datasources.iri.iri_seasonal_forecast
-   :members:
-   :undoc-members:
-   :show-inheritance:
+
+Probability forecast
+""""""""""""""""""""
+
+.. autoclass:: aatoolbox.IriForecastProb
+   :inherited-members:
+
+Dominant forecast
+"""""""""""""""""
+
+.. autoclass:: aatoolbox.IriForecastDominant
+   :inherited-members:
 
 NDVI (USGS eMODIS)
 ^^^^^^^^^^^^^^^^^^
 
-Products
+.. automodule:: aatoolbox.datasources.usgs.ndvi_products
+
+Smoothed
 """"""""
 
-.. automodule:: aatoolbox.datasources.usgs.ndvi_products
-   :members:
-   :undoc-members:
-   :show-inheritance:
+.. autoclass:: aatoolbox.UsgsNdviSmoothed
+   :inherited-members:
 
-Base class
-""""""""""
+Percent of median
+"""""""""""""""""
 
-.. automodule:: aatoolbox.datasources.usgs.ndvi_base
-   :members:
-   :undoc-members:
-   :show-inheritance:
+.. autoclass:: aatoolbox.UsgsNdviPctMedian
+   :inherited-members:
 
+Median anomaly
+""""""""""""""
+
+.. autoclass:: aatoolbox.UsgsNdviMedianAnomaly
+   :inherited-members:
+
+Difference from previous year
+"""""""""""""""""""""""""""""
+
+.. autoclass:: aatoolbox.UsgsNdviYearDifference
+   :inherited-members:
 
 Utilities
 ---------
 
+GeoboundingBox
+^^^^^^^^^^^^^^
+
+.. automodule:: aatoolbox.utils.geoboundingbox
+.. autoclass:: aatoolbox.GeoBoundingBox
+   :members:
+
 Raster module
 ^^^^^^^^^^^^^^^^
 
-The raster module provides accessor utilities for xarray
-data arrays and datasets accessible using the `aat` accessor.
+.. automodule:: aatoolbox.utils.raster
 
 Data arrays
 """""""""""
 
 .. autoclass:: aatoolbox.utils.raster.AatRasterArray
-    :members:
-    :undoc-members:
-    :show-inheritance:
+   :members:
+   :inherited-members: RasterArray
 
 Datasets
 """"""""
+
 .. autoclass:: aatoolbox.utils.raster.AatRasterDataset
-    :members:
-    :undoc-members:
-    :show-inheritance:
+   :members:
+   :inherited-members: RasterDataset
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -1,8 +1,11 @@
 AA Toolbox API
 ==============
 
+Data sources
+------------
+
 CHIRPS
-------
+^^^^^^
 
 .. automodule:: aatoolbox.datasources.chirps.chirps
    :members:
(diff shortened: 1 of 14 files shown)


docs/source/utilities/raster.rst
```diff
@@ -0,0 +1,58 @@
...
+RasterDataset extensions respectively. This ensures
+cleaner code in the module as ``rio`` methods are
+available immediately, but also means a couple of
+design decisions are followed.
+
+The ``xarray.DataArray`` and ``xarray.Dataset``
+extensions here inherit from rioxarray base classes.
+Thus, methods that are identical for both objects
+are defined in a mixin class ``AatRasterMixin`` which
+can be inherited by the two respective extensions.
```
turnerm, line 27: I think the code implementation details can be kept to the source code, in the docs I would just focus on the usage (although indeed the complicated design does have impact on the functionality, so keep whatever you think is needed)

docs/source/utilities/raster.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Usage
+-----
+
+To make the extension accessible, you only need to import
+``aatoolbox`` directly. As a simple example, we can create
+a simple data array that has incorrect coordinates
```
turnerm, line 34: Already used "simple" in the sentence

docs/source/utilities/raster.rst
```diff
@@ -0,0 +1,58 @@
...
+    da = xarray.DataArray(
+      numpy.arange(16).reshape(4,4),
+      coords={"lat":numpy.array([87, 88, 89, 90]),
+              "lon":numpy.array([70, 69, 68, 67])}
+    )
+    da_inv = da.aat.invert_coordinates()
+
+    # check they have inverted
+    da_inv.get_index("lon")
+    da_inv.get_index("lat")
```
turnerm, line 54: Can you print out the results, before and after, to show the change?

docs/source/utilities/raster.rst
```diff
@@ -0,0 +1,58 @@
...
+
+As above, for any data array (or
+dataset) ``da``, the extension is available for use
+directly using ``da.aat.method_or_property``.
```
turnerm, line 58: I think it would be valuable to add `compute_raster_stats` to the docs here, as I'd imagine that's one of the most useful features

turnerm (commented):
Thanks so much for tackling this one, I think it will be so great to have this functionality available! Just some small changes suggested. 

One other thing though (sorry, as always). What do you think of renaming USGS to NDVI in the list of data sources? My feeling is that if someone is scanning the list, they would recognize NDVI as something they know. We can rename "datasources" to "data sets" if that would make more sense. Let me know your thoughts. (Maybe this is too much for this PR, we can always go back to the discussion on #42)

docs/source/utilities/raster.rst
caldwellst: Good point. I read through and don't think it has any info the user actually wants, because really they just care about how to access it not that it's an extension of `rioxarray` for instance which they'd probably access anyway using `rio.method()`

docs/source/utilities/raster.rst
caldwellst: :O

caldwellst:
Okay interesting point on the modules. Have commented on the other discussion about the module structuring but as you say, I think out of scope of this PR. Hope the rest of the changes are good!

caldwellst:
Added some documentation the raster module on the shifting of coordinate ranges. However, I realised that I don't think the API documentation is generated correctly for the raster module. I tried adding it to `__all__` in the aatoolbox initiation, thinking the module would then be a member and methods available in the API. However, that hasn't worked. @turnerm any ideas?

Also beyond the scope of this PR maybe, but can make an issue if you agree, should we have direct links to the API of a module from their main documentation page? That might make it easier for users to directly access the full module documentation directly from the data source page they are most interested in.

docs/source/datasources/usgs_ndvi.rst
```diff
@@ -1,5 +1,5 @@
 USGS
```
turnerm, line 1: What I meant with my comment is if you could change this title to NDVI instead of USGS -- that way "NDVI" would show up in the list of data sets instead of USGS (so a bit more descriptive for people who are trying to see what's available). But no worries if you disagree and we can keep discussing in the relevant issues

docs/source/utilities/raster.rst
```diff
@@ -0,0 +1,90 @@
...
+    da_inv = da.aat.invert_coordinates()
+
+    # check they have inverted
+    da_inv.get_index("lon")
+    #> Int64Index([67, 68, 69, 70], dtype='int64', name='lon')
+    da_inv.get_index("lat")
+    #> Int64Index([90, 89, 88, 87], dtype='int64', name='lat')
+
+One of the more useful functionalities of the module
+is providing an easy ability to calcuate raster statistics
```
turnerm, line 48: Can you turn this section (and the one above) into sub-sub-sections, calling this one "raster statistics"? I just want to emphasize the functionality as I think it's something to really advertise!!

turnerm (approved):
> Added some documentation the raster module on the shifting of coordinate ranges. However, I realised that I don't think the API documentation is generated correctly for the raster module. I tried adding it to __all__ in the aatoolbox initiation, thinking the module would then be a member and methods available in the API. However, that hasn't worked. @turnerm any ideas?

It shows up for me! You have to make sure that you have the latest `pip install .` (or `pip install -e .`), and then run the doc generation command which is specified in CONTRIBUTING.rst. But let me know if it still doesn't work. 

Looks ready to merge, just a couple of last tiny comments!

docs/source/datasources/usgs_ndvi.rst
caldwellst: No that totally makes sense, I didn't understand exactly what you meant!

turnerm:
Ah oops I read your question too quickly, I now realize you were referring to the API docs, indeed `raster` also does not appear for me there too. Did you manage to solve it?

docs/source/datasources/usgs_ndvi.rst
caldwellst: I forgot to change in here but have put in the PR directly about making clear the issues with USGS NDVI data.

caldwellst:
No I forgot about that and will need to explore what that issue is. Checking it out now.

caldwellst:
Not sure why not auto-documenting, but something to do with there being an issue of considering it part of the module. However, @turnerm, I have proposed a new way of structuring the API. It required being more explicit about the structure, but it means that there are easy navigable headers to data sources for instance.

docs/source/api.rst
```diff
@@ -1,5 +1,84 @@
...
 ==============
 
-.. automodule:: aatoolbox
+Data sources
+------------
+
+CHIRPS
+^^^^^^
+
+.. automodule:: aatoolbox.datasources.chirps.chirps
```
turnerm, line 10: Do you know if it's possible to add the items from `aatoolbox` directly, i.e. instead of `aatoolbox.datasources.chirps.chirps`, just `aatoolbox.chirps`? I can't get it to work. My concern is now that the full import path shows up it could be a bit confusing. But it's not a big deal.

docs/source/api.rst
```diff
@@ -1,5 +1,84 @@
...
+   :members:
+   :undoc-members:
```
turnerm, line 12: Maybe it's not necessary to include `undoc-members`? Not sure we want to generate docs for functions with docstrings. (On a similar note, not sure we need `show-inheritance`, tbh I'm not sure what it does exactly)

docs/source/api.rst
```diff
@@ -1,5 +1,84 @@
...
+
+Products
+""""""""
+
+.. automodule:: aatoolbox.datasources.usgs.ndvi_products
    :members:
+   :undoc-members:
+   :show-inheritance:
+
+Base class
```
turnerm, line 53: I don't think we need to include this

docs/source/api.rst
```diff
@@ -1,5 +1,84 @@
...
+
+.. automodule:: aatoolbox.datasources.fewsnet.fewsnet
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+IRI
+^^^
+
+Seasonal forecasts
```
turnerm, line 34: Maybe remove this subheading since there is only one IRI forecast for now anyway

docs/source/api.rst
```diff
@@ -1,5 +1,84 @@
 AA Toolbox API
 ==============
 
-.. automodule:: aatoolbox
```
turnerm, line 4: Country config and geo bounding box are missing

turnerm (commented):
@caldwellst I really like that change, great idea! Indeed it is a bit annoying to have to add everything by hand but not prohibitively so, and I think the gains are worth it. Just left a few comments.

docs/source/api.rst
caldwellst: Found how to do what I wanted, show the inherited methods, in another way.

caldwellst:
Okay, so it's now even slightly more complex to do the documentation, but I don't think is too onerous given the amount of sources we have and makes for cleaner structure. What we want to do is essentially:

1. Document each module in the `.py` file itself, which we can then bring in using `automodule`, referring to its full path. We don't document any children or anything else here.
2. Below that, we can document classes or functions as necessary. For classes that directly inherit from `DataSource` or other low-level base classes, we just document their `:members:`.
3. For classes that inherit from a base class that defines specific methods, like `ChirpsDaily` from `_Chirps`, we specify `:inherited-members:` because there are no public methods we want documented specifically within `ChirpsDaily`.
4. For something like the raster modules, which inherit methods from a base class but the data array and dataset classes have unique methods at points, we specify both `:members:` and `:inherited-members:` while also excluding documentation of the base rioxarray classes `RasterArray` and `RasterDataset` by explicitly listing them.

turnerm (approved):
Wow this looks awesome, I think it's a great improvement, thanks Seth!

Last couple of things (but you can merge without me seeing them they are very small):

1. Could you add the last blurb that you wrote as instructions to CONTRIBUTING.rst?
2.  Could you sort the `__all__` list in `src/aatoolbox/__init__.py`, I'm undecided if it's better to sort it in the order of import, or in the same order as `api.rst`, you can choose, aha.

Issue #24, opened 2021-11-03
Specifying correct Python version when running pytest with pyenv


caldwellst:
When testing the `pytest` framework, I ran into an issue where `pytest` didn't recognise the `--cov` argument passed to it. Below is a small example:

```shell
>> pytest --cov=src
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov=src
  inifile: None
  rootdir: /Users/caldwellst/Desktop/temp
```

Turns out users of `pyenv` could run into this issue where `pytest` runs off a different version of Python than what is active through `pyenv` (and that version in this case didn't have `pytest-cov` installed). Seems that there has been a fall through in my case with the shims, not sure why, so have to explicitly call `pytest` from my active Python version.

```shell
>> python -m pytest --cov=src
```
This is all based off of [this SO post](https://stackoverflow.com/questions/40718770/pytest-running-with-another-version-of-python). Raising this here in the case we wanted to somehow add this to documentation. It might be too niche to include, but wanted to at least have an issue active for others that may run into it.

turnerm:
@caldwellst thanks for flagging this, could you add a couple of lines about it to CONTRIBUTING.rst and then close?

PR #151, opened 2022-11-14
Address #24

caldwellst:
Adds short info about potential errors in pytest running from the incorrect environment.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -43,6 +43,15 @@ To run the tests and view coverage, execute:
 
    pytest --cov=aatoolbox
 
+This code may run the testing using Python or environment different
+to the testing environment, which may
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -43,6 +43,15 @@ To run the tests and view coverage, execute:
 
    pytest --cov=aatoolbox
```
turnerm, line 44: I would just swap out this command with the one you specified below, you can then give a warning that "If you run it directly with pytest, you may get the following error..." or something

turnerm (approved):
Thanks so much for taking care of this so quickly! Just one small suggestion

CONTRIBUTING.rst
caldwellst: Good idea! Addressed!

caldwellst:
Okay, see what you think of the new text!

turnerm:
That's perfect, good to merge any time!

Issue #150, opened 2022-11-14
NDVI data removed from USGS portal


caldwellst:
In an exciting development, NDVI data has been entirely removed from the [USGS portal](https://earlywarning.usgs.gov/fews) due to significant issues with the MODIS sensor.

> Important Notice: Due to severe degradation of data quality being provided by the AQUA satellite, which carries the MODIS sensor, we have suspended access to eMODIS NDVI products until further notice. In addition, our evapotranspiration (ETa) modeling relies on MODIS derived land surface temperature products which have been degrading in quality since May of 2022. Therefore, we have chosen to suspend access to all ETa products that include data from May forward. We apologize for the inconvenience.

We will somehow need to quarantine our module lol? Maybe temporarily adding a clear warning (whatever the highest level of logging is) when the class is instantiated about the quality of data from May 2022 (if the dates of the class extend beyond that) and that no new downloading is available, an error message repeating the same thing for download, but leaving the class up for processing and loading previously downloaded data.

caldwellst:
Issue with module addressed in #154, closing and will address additional sources in #153.

PR #154, opened 2022-11-15
Add warnings to USGS NDVI data, addresses #150

caldwellst:
Addresses adding warnings based on #150.

### DIFF
--- a/docs/source/datasources/usgs_ndvi.rst
+++ b/docs/source/datasources/usgs_ndvi.rst
@@ -1,5 +1,15 @@
-USGS
-======
+NDVI (USGS eMODIS)
+==================
+
+Warning
+-------
+
+The MODIS sensor has been reported by USGS to
+have degraded in quality since May 2022 (dekad 13), and
+updates to this data source have stopped. This module
+remains for users to have access to historic data but
+recent data is unavailable and care should be used
+analyzing any data since dekad 13 of 2022.
 
 Background
 ----------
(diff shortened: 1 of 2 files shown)


turnerm (commented):
Looks great! My only comment would be that in CHRIPS and now also GloFAS, we don't just give a warning with an out of bounds date, but throw an error. Not sure if you think it's worth being consistent across these. (Open to changing CHIRPS and GloFAS if you think warning is the better way to go)

caldwellst:
Ooh that's an interesting point. I think that could be a good option here as well. I think the only potential annoyance to users is that for dekadal or pentadal data, it's a bit less clear where the bounds are. They might just want everything up to a certain date but just specify to end of year to make sure they don't miss a specific time or dekad?  So like, where do we set the "bounds" in this instance? Last day of the last available dekad?

caldwellst:
@turnerm @castledan Thoughts?

castledan:
What happens in the code now? Do you raise a warning and adjust the end date to the last available one?

caldwellst:
Ah, so I'm realising a bit of an artifact of the NDVI data, but there are random dekads missing from the data in the online portal due to random errors from the sensor. Because of this, I couldn't throw an error if dekadal data unavailable, but simply log it in info that no data available for that year and dekad. Now I throw an explicit warning if the end date is past a certain point since there's more systematic missing data, but just keep the end date to avoid hardcoding anything and allow the logging info pass on messages about data not available.

Issue #136, opened 2022-10-14
Nightly build


turnerm:
Run the tests nightly to catch any package update issues

PR #156, opened 2022-11-16
Build and test nightly, address #136

caldwellst:
Having fun trying to tackle various things and learning, so hope these are helpful!


turnerm (approved):
LGTM!

PR #157, opened 2022-11-16
Add python 3.11 to testing, address #142

caldwellst:
Adding 3.11 to testing. Let's see if any specific issues pop up during the test runs that need addressing.


turnerm (approved):
Amazing, thanks!

PR #161, opened 2022-11-24
Python req upgrade

caldwellst:
Okay, have gone through and upgraded to Python 3.11 for our dev environment. Let me know if I need to change more, I just reconstructed the `requirements.txt` files and adjusted `CONTRIBUTING.rst`.


requirements/requirements-dev.txt
```diff
@@ -16,7 +16,7 @@ appdirs==1.4.4
     # via
```
turnerm, line 16: Python version should be changed here

turnerm (approved):
Looks amazing! I don't think it needs to go in the changelog.

Issue #147, opened 2022-11-11
Fix Ethiopia config file


turnerm:
Being addressed in PR #125

PR #162, opened 2022-11-25
Update COD AB config file customization

turnerm:
Addresses #147, and supersedes #125. 

Make it possible to specify custom COD AB layer names in the configuration file.

### DIFF
--- a/tests/test_country_config.py
+++ b/tests/test_country_config.py
@@ -95,34 +95,6 @@ def test_codab_validate_layer_base_name(mock_parse_yaml):
 
 
 def test_codab_validate_admin_level_max(mock_parse_yaml):
-    """Test that layer basename requires correct placeholder."""
-    config_base = {
-        "iso3": "abc",
-        "codab": {
-            "hdx_dataset_name": "fake_dataset_name",
-            "layer_base_name": "layer_base_name_{admin_level}",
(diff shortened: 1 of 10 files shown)


tests/test_country_config.py
```diff
@@ -94,7 +94,63 @@ def test_validate_codab_layer_base_name(mock_parse_yaml):
         create_country_config(iso3="abc")
 
 
-def test_validate_fewsnet_region_name(mock_parse_yaml):
+def test_codab_validate_admin_level_max(mock_parse_yaml):
+    """Test that layer basename requires correct placeholder."""
```
caldwellst, line 98: It looks like below you're only testing for the `admin_level` not actually about the placeholder. I think it's accidentally the same base code as the next test.

caldwellst (changes requested):
Hey Monica! Just a couple of changes, not much at all. Mainly got slightly confused because I had found an incorrect (or not CODAB) geo boundaries on HDX for Nepal that had ADM3 ha.

tests/test_country_config.py
turnerm: Oops, some code copy-pasting gone terribly awry, fixed

caldwellst (approved):

Issue #149, opened 2022-11-11
Incorrect IRI auth downloads bogus files


turnerm:
If the IRI auth is incorrect, the request still returns 200, and the toolbox will download an HTML file instead of the expected NetCDF.

This should not be allowed to happen, what we can do is check the requests headers [here](https://github.com/OCHA-DAP/pa-aa-toolbox/blob/main/src/aatoolbox/datasources/iri/iri_seasonal_forecast.py#L203) and throw an error if they are not as expected.

turnerm:
Closed by #163

PR #163, opened 2022-11-25
Fix IRI auth

turnerm:
Addresses #149. Check request headers to verify IRI auth, instead of downloading broken files.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Fixed
 
 - The check in ``DataSource`` for the required configuration file
   section now also checks if the section is ``None``
+- IRI download method now checks request headers to verify authentication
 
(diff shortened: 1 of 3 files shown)


caldwellst (approved):
Looks good to me! My one thought when reading it is, is there any risk in raising the exception every time something other than a NetCDF file is returned? Like, could it get raised due to bad connection? Is there a message returned in the response we could directly pass back to the user or a consistent error returned about lack of authentication we could check for instead?

I think good to merge as is just something that came to mind!

turnerm:
Good point Seth. I didn't want to speculate about all the possible ways that things could go wrong, so I prefer just checking that they go right, but I think in that case I should make the exception message more broad.  Will do that and merge.

PR #173, opened 2023-01-29
Fix typo in LICENSE link

caldwellst:
Typo in LICENSE link we missed previously.


turnerm (approved):
Thanks so much Seth!

turnerm:
@caldwellst don't forget to update the changelog

caldwellst:
Ah okay. I didn't update because none of this matters to the end user, but will also put in there then!

PR #175, opened 2023-02-03
Drop 3.7 

turnerm:
The main change is to drop Python 3.7 since `hdx-python-utils` has dropped it (EOL is in a few months anyway), and the tests keep failing. 

Other small changes:
- Update the requirements
- Update the pre-commit packages
- Remove `sphinx-rtd-theme` since I didn't like it that much anyway and always has outdated dependencies

I originally wanted to add 3.12 because I saw that Mike added it but then I realized that the stable release doesn't come out until October so I think it can wait!


caldwellst (approved):
Looks good to me! Nothing I could find missing related to 3.7. Just remember the CHANGELOG. Should you merge to develop, and then together with the other unreleased changes merge to main as 1.0.1?

PR #187, opened 2023-03-22
Bugfix/set spatial dims

### LINKED ISSUE #183: raster module set spatial dims
For the Yemen flooding work, I was using the raster `clip_box` function. However, I found that I had to set the spatial dims in `rio` first or else it wouldn't work, see the code [here](https://github.com/OCHA-DAP/pa-aa-yem-flooding/blob/main/src/datasource_extensions.py#L68-L71). 

Could the spatial dim setting be done when initializing `oap`?

caldwellst:
Fixes #183.

### DIFF
--- a/src/ochanticipy/utils/raster.py
+++ b/src/ochanticipy/utils/raster.py
@@ -54,17 +54,13 @@ class OapRasterMixin:
 
         # Adding lat/lon to set of default spatial dims
         if "lat" in self._obj.dims and "lon" in self._obj.dims:
-            self._x_dim = "lon"
-            self._y_dim = "lat"
-            self._obj.rio._x_dim = "lon"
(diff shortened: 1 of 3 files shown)


src/ochanticipy/utils/raster.py
```diff
@@ -56,11 +56,15 @@ def __init__(self, xarray_obj):
         if "lat" in self._obj.dims and "lon" in self._obj.dims:
             self._x_dim = "lon"
             self._y_dim = "lat"
+            self._obj.rio._x_dim = "lon"
+            self._obj.rio._y_dim = "lat"
```
turnerm, line 60: you can make these one line, like `self._x_dim = self._obj.rio_x_dim = "lon"`

turnerm (approved):
LGTM! :)

PR #188, opened 2023-03-22
Add easy HRP countries

turnerm:
Here are the "easy" HRP countries (without strange COD AB issues mentioned [here](https://github.com/OCHA-DAP/ocha-anticipy/issues/181). 

Opening as a draft because I still have to test everything myself!


src/ochanticipy/config/countries/lby.yaml
```diff
@@ -0,0 +1,7 @@
+iso3: lby
+codab:
+  hdx_dataset_name: lby_adm_unosat_lbsc_20180507_SHP.zip
+  layer_base_name: lby_adm_unosat_lbsc_20180507_SHP/lby_admbnda_adm{admin_level}_unosat_lbsc_20180507.shp
+  admin_level_max: 2
+usgs_ndvi:
+  area_name: east-africa
```
caldwellst, line 7: Should be North Africa.

caldwellst (approved):
I think it looks good! Just one minor change, plus my question about just doing the `iso2` change in this PR? Wondering, is there any way we should be testing these to make sure they don't silently fail eventually?

turnerm:
> Wondering, is there any way we should be testing these to make sure they don't silently fail eventually?

Yes this will undoubtedly be an issue, the solution IMO would be some kind of E2E testing. But in the meantime we can just monitor the COD Skype group for any updates.

src/ochanticipy/config/countries/lby.yaml
turnerm: Oops thanks!

docs/config.rst
```diff
@@ -15,12 +15,32 @@ and we are continuously adding more.
 
 .. _list of supported countries:
 
-A list of countries that we currently support:
+We currently support 20 out of 25 humanitarian response plan (HRP) countries:
```
caldwellst, line 18: I think here should we just add a caveat that not all data is available for all countries? Cause people might read it and expect their country to be there.

caldwellst (approved):
Just one question about adding a caveat that not all data sources available across all countries when listing "supported countries". Looks great!

Issue #106, opened 2022-04-19
Create dekadal class


caldwellst:
@turnerm had a very nice idea from the NDVI PR:

> In general I think all the functions in here make more sense as a Dekad class.
>
> It can be out of scope for this PR, but I'm wondering if you could consider [subclassing datetime.date](https://www.aaronoellis.com/articles/subclassing-datetime-date-in-python-3).
>
> I wasn't able to find any Python package that really handles dekads, so this could be a good opportunity to create something useful, maybe even as a separate package.

Putting here because I think out of scope for this PR but definitely something we should look into.

caldwellst:
Followed the advice for a separate package and created [kalendar](https://github.com/OCHA-DAP/kalendar). Once reviewed and we are comfortable with the release, will switch the dekad implementation in this project to use `kalendar`.

PR #189, opened 2023-03-23
Implement kalendar module

caldwellst:
Implements #106.

I brought in the `kalendar` library into our `dates.py` module. I think we still need our own module to allow for the broader implementation of the user inputted ISO dates/strings and wider work, but has made the implementation much simpler.

This has all been used on the backend for NDVI. However, I made the decision NOT to include the dekad class on the end output. I think that would make it confusing for end users if they potentially saw classes they weren't familiar with or had no desire to use, so rather kept the format of output files having year, dekad, and date columns.

Look forward to your review and thoughts!

### DIFF
--- a/src/ochanticipy/datasources/usgs/ndvi_base.py
+++ b/src/ochanticipy/datasources/usgs/ndvi_base.py
@@ -28,27 +28,22 @@ import geopandas as gpd
 import pandas as pd
 import rioxarray  # noqa: F401
 import xarray as xr
+from kalendar import Dekad
 from rasterio.errors import RasterioIOError
 
 import ochanticipy.utils.raster  # noqa: F401
 from ochanticipy.config.countryconfig import CountryConfig
 from ochanticipy.datasources.datasource import DataSource
-from ochanticipy.utils.dates import (
-    compare_dekads_gt,
-    compare_dekads_lt,
-    dekad_to_date,
-    expand_dekads,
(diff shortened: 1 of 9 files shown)


src/ochanticipy/datasources/usgs/ndvi_base.py
```diff
@@ -102,15 +98,19 @@ def __init__(
         self._data_variable_suffix = data_variable_suffix
 
         # set dates for data download and processing
-        self._start_date = get_dekadal_date(
-            input_date=start_date, default_date=_EARLIEST_DATE
+        self._start_date = get_kalendar_date(
```
turnerm, line 101: Why didn't you keep `get_dekadal_date` here?

turnerm (commented):
Hey Seth, this is cool, I didn't know about `@overload`! I was reading a bit and I think the preferred way to do this in Python (which maybe could be applied to some other methods, also in Kalendar -- sorry I should have thought of this earlier!) is to use dispatching, see for example [this article](https://martinheinz.dev/blog/50). Do you think it could work for this case?

src/ochanticipy/datasources/usgs/ndvi_base.py
caldwellst: hahaha, because I am a dingus. I didn't have `get_dekadal_date()` at first.

PR #193, opened 2023-03-28
Feature/add hrp countries

caldwellst:
So, I have put together the setup for the additional HRP countries. The only two countries missing are:

- Afghanistan, since they don't have shape files available.
- Myanmar, because they do not have a COD AB (so I think is fair we don't include).


src/ochanticipy/config/countryconfig.py
```diff
@@ -7,37 +7,59 @@
...
     admin2_name: Optional[str]
     admin3_name: Optional[str]
     admin4_name: Optional[str]
-    custom_layer_names: Optional[list]
+    custom_layer_names: Optional[List[CustomLayerNames]]
+
+    @validator("hdx_resource_name")
+    def _validate_hdx_resource_name(cls, hdx_resource_name, values):
+        """Ensure hdx_resource_name is str or list for all admin areas."""
+        if not isinstance(hdx_resource_name, str):
```
turnerm, line 54: Maybe just check if it is a list, that would make it easier to follow the logic

src/ochanticipy/config/countryconfig.py
```diff
@@ -7,37 +7,59 @@
 from ochanticipy.utils.io import parse_yaml
 
 
+class CustomLayerNames(BaseModel):
+    """Custom layer names for CodAB."""
+
+    name: str
+    hdx_resource: Optional[int]
```
turnerm, line 14: Overall this is a great implementation if we ever need it, however I would say since we don't have a situation where a country has both multiple resources and custom layer names, we can remove it for now just to keep it simple and avoid over engineering. And if it does come up then we have this code in our back pockets!

src/ochanticipy/datasources/codab/codab.py
```diff
@@ -30,12 +31,30 @@ def __init__(self, country_config: CountryConfig):
...
+            f"{self._country_config.iso3}_{self._datasource_base_dir}"
         )
 
-    def download(self, clobber: bool = False) -> Path:
+        # account for files stored across multiple resources
+        res_name = self._datasource_config.hdx_resource_name
+        self._multiple_resources = isinstance(res_name, list)
+
+        if self._multiple_resources:
+            self._resource_dirs = [
```
turnerm, line 43: You don't need to make this an instance variable if it's not used outside of the method

src/ochanticipy/datasources/codab/codab.py
```diff
@@ -56,12 +75,16 @@ def download(self, clobber: bool = False) -> Path:
...
+        for filepath, hdx_resource_name in zip(
+            self._raw_filepaths, self._hdx_resource_list
+        ):
+            self._download(
+                filepath=filepath,
+                hdx_dataset=f"cod-ab-{self._country_config.iso3}",
+                hdx_resource_name=hdx_resource_name,
+                clobber=clobber,
+            )
+        return self._base_dir
```
turnerm, line 87: In the case of multiple files, I try to return the full path to the directory containing the files

src/ochanticipy/datasources/codab/codab.py
```diff
@@ -147,38 +171,50 @@ def load_custom(self, custom_layer_number: int = 0) -> gpd.GeoDataFrame:
...
                 f"{custom_layer_number}th custom layer requested but not "
                 f"available in {self._country_config.iso3.upper()} config file"
             ) from err
-        return self._load_admin_layer(layer_name=layer_name)
+        return self._load_admin_layer(
+            layer_name=layer_name, admin_level=admin_level
+        )
+
+    def _load_admin_layer(
+        self, layer_name: str, admin_level: Optional[int] = None
```
turnerm, line 191: I think we should just make admin level default 0 in the `load` method, I've been meaning to change that anyway for awhile

turnerm (commented):
Hey Seth, thanks for all the work on this, it looks really great. I think my main comments are:

- remove the custom layer name / multiple resources capability for now - I think custom layer names are very rare, hopefully we won't ever need it (plus HAPI is coming soon anyway)
- update the COD AB documentation to reflect the new variable names and explain how to implement multiple resources 
- this is a larger one, but you can also ignore it: I would make the implementation in codab.py a bit more duck-typing-y. Like I would not explicitly check the types, but rather have `download` and `process` try to handle either case. Hope that makes sense. 

And don't forget to update the changelog!

src/ochanticipy/datasources/codab/codab.py
caldwellst: It's used in the `_load_admin_layer()`

src/ochanticipy/datasources/codab/codab.py
caldwellst: In this case, I think we are doing that. This should return the base directory where all of the zip files are stored. Since we don't unzip any of the files (if multiple), `self._base_dir` should be the directory that contains all of the files (the zip files). Should I change it to be the path to the zip file if a single file is downloaded, and the directory if multiple?

turnerm: [here](https://github.com/OCHA-DAP/ocha-anticipy/pull/193/files#diff-bab90d7c88ef7e9d8c3300c7c71b4733fb7c1312252adefdb2481416ab892991R34) is where you set `base_dir`, it only ends up being `{iso3}_cod_ab`

> Should I change it to be the path to the zip file if a single file is downloaded, and the directory if multiple?

You could but I wouldn't bother if it's too complicated

docs/datasources/codab.rst
```diff
@@ -75,35 +75,39 @@ should be setup as follows:
...
-      custom_layer_names:
-        - npl_admbnda_districts_nd_20201117
+        hdx_resource_name:
+            - nga_adm_osgof_20190417.zip
+            - nga_additional.zip # contrived example for documentation
+        layer_base_name: nga_admbnda_adm{admin_level}_osgof_20190417.shp
+        admin_level_max: 3
+        adm1_name: nga_adm1.shp # contrived example for documentation
+        custom_layer_names:
+            - name: nga_admbnda_senDist_inec_osgof_20190417.shp
```
turnerm, line 85: You can remove the "name" part

src/ochanticipy/config/countries/nga.yaml
```diff
@@ -1,11 +1,11 @@
 iso3: nga
 codab:
-  hdx_dataset_name: nga_adm_osgof_20190417.zip
+  hdx_resource_name: nga_adm_osgof_20190417.zip
   layer_base_name: nga_admbnda_adm{admin_level}_osgof_20190417.shp
   admin_level_max: 3
   # Nigeria also has districts that don't fall into a particular admin level
   custom_layer_names:
-    - nga_admbnda_senDist_inec_osgof_20190417.shp
+    - name: nga_admbnda_senDist_inec_osgof_20190417.shp
```
turnerm, line 8: Remove the "name" here

src/ochanticipy/config/countries/som.yaml
```diff
@@ -1,13 +1,13 @@
...
-  hdx_dataset_name: som_adm_ocha_itos_20230308_SHP.zip
+  hdx_resource_name: som_adm_ocha_itos_20230308_SHP.zip
   layer_base_name: som_admbnda_adm{admin_level}_ocha_20230308.shp
   admin_level_max: 2
   # Somalia also has states and 2 operational zones
   custom_layer_names:
-    - som_admbnda_state_ocha_20230308.shp
-    - som_admbnda_opz1_ocha_20230308.shp
-    - som_admbnda_opz2_ocha_20230308.shp
+    - name: som_admbnda_state_ocha_20230308.shp
```
turnerm, line 8: Remove the "names"

src/ochanticipy/config/countryconfig.py
```diff
@@ -7,37 +7,61 @@
 from ochanticipy.utils.io import parse_yaml
 
 
+class CustomLayerNames(BaseModel):
```
turnerm, line 10: Can remove this class, that way you won't need "name" in front of the custom layer names which is redundant

turnerm (approved):
Great thanks so much Seth, this is great! We are HRP complete! :D Only two small things:
 - the returned path, it's not quite the right one
 - removing `CustomLayerNames` class and adjusting the documentation

PR #196, opened 2023-04-17
Hotfix/raster doctests

caldwellst:
Simple fix to raster doctests addressing the failing tests. The tests are not failing due to any functional issues, just that the docstrings reference ``Int64Index`` from pandas which has now been r[emoved and replaced by ``Index``](https://stackoverflow.com/questions/71083866/pandas-int64index-fix-for-futurewarning), so the doctests were failing. Because of this, just figured we can release alongside the rest of develop since many changes are coming, rather than hot fixing directly to main.


turnerm (approved):
I like your optimism. Thanks for fixing this!

Issue #180, opened 2023-03-21
Misc bugs mostly GloFAS


turnerm:
Some small things to fix that I found while working with GloFAS data:

- related to #179, give a warning if users have more than 500 requests, advise them to split up into two instances
- add "Sleeping for 1 minute.." when that is happening during the download queue
- `check_file_existence` shouldn't exclusively use the terminology "downloading" because it's also applied to processing
- bug in the final list of commands of GloFAS docs
- set COD AB default to be admin 0 (as is stated in the documentation)

PR #199, opened 2023-05-29
Small fixes

turnerm:
Addresses #180

### DIFF
--- a/src/ochanticipy/datasources/codab/codab.py
+++ b/src/ochanticipy/datasources/codab/codab.py
@@ -71,13 +71,13 @@ class CodAB(DataSource):
         """
         logger.info("`process()` method not implemented for CodAB.")
 
-    def load(self, admin_level: int) -> gpd.GeoDataFrame:  # type: ignore
+    def load(self, admin_level: int = 0) -> gpd.GeoDataFrame:  # type: ignore
         """
         Get the COD AB data by admin level.
 
         Parameters
         ----------
-        admin_level: int
+        admin_level: int, default = 0
             The administrative level
 
         Returns
--- a/src/ochanticipy/datasources/glofas/glofas.py
+++ b/src/ochanticipy/datasources/glofas/glofas.py
@@ -20,6 +20,7 @@ from ochanticipy.utils.geoboundingbox import GeoBoundingBox
 _MODULE_BASENAME = "glofas"
 _HYDROLOGICAL_MODEL = "lisflood"
 _RIVER_DISCHARGE_VAR = "dis24"
+_CDS_MAX_REQUESTS = 500
 _REQUEST_SLEEP_TIME = 60  # seconds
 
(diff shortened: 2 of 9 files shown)


src/ochanticipy/datasources/codab/codab.py
```diff
@@ -71,13 +71,13 @@ def process(self, *args, **kwargs):
...
 
-    def load(self, admin_level: int) -> gpd.GeoDataFrame:  # type: ignore
+    def load(self, admin_level: int = 0) -> gpd.GeoDataFrame:  # type: ignore
         """
         Get the COD AB data by admin level.
 
         Parameters
         ----------
-        admin_level: int
+        admin_level: int, default = 0
```
caldwellst, line 80: Nice. I made changes in the `feature/add_hrp_countries` branch but have changed to match this so we don't have to resolve later.

src/ochanticipy/datasources/glofas/glofas.py
```diff
@@ -138,6 +139,15 @@ def __init__(
             dtstart=self._start_date,
             until=self._end_date,
         )
+        if self._date_range.count() > _CDS_MAX_REQUESTS:
```
caldwellst, line 142: So, question, could an eventual implementation make the maximum requests and then sleep before making the additional ones, preventing the user from having to create two separate instances if they want to analyse data across a time that contains more points than `_CDS_MAX_REQUESTS`?

caldwellst: Not saying to be done now, but should it be an issue or no?

src/ochanticipy/utils/check_file_existence.py
```diff
@@ -65,7 +65,7 @@ def check_file_existence(
     # check filepath exists -> clobber
     usage_dict = {
         True: {True: "overwriting existing", False: "using existing"},
-        False: {True: "downloading new", False: "downloading new"},
```
caldwellst, line 68: Good catch.

caldwellst (approved):
All looks good to me! Just a couple of questions added, but nothing critical to merging

src/ochanticipy/datasources/glofas/glofas.py
turnerm: Oh yeah definitely, good question. It's actually how I started writing this module in the first place, using `asyncio` and a producer / consumer architecture. Then I realized that the request limit was very high and we wouldn't reach it (at the time, this was a few months ago...) so I dropped this plan, as `asyncio` makes things more complicated and testing hard. 

But _now_ with the forecast we do reach the queue limit if asking for all historical available forecasts, oops. So indeed this is a temporary fix.

turnerm: Ah yes  and it should be an issue, let me add it

PR #205, opened 2023-08-16
GloFAS new version

turnerm:
Update to allow both version 3 and 4 in GloFAS. Some things to note:

- The directory structure is the same but the filenames will now have the version number in them. This means that any downloads done with previous AnticiPy versions will now have broken filepaths. If you want to work with them and the latest AnticiPy then you will either need to fix the filenames by hand or re-download. 
- The main annoying thing is the differences in the versions is in the reforecast. The year range for version 4 is shifted up, and only the months March to August are available. (Does anybody know if this will change in the future?). I've tried to account for this but it could be helpful to test in case something wonky comes up. 
- Even though the resolution has changed, I've kept the old bounding box rounding to x.x5. It's not clear any more from the CDS API what is now recommended here, I should probably ask Ervin. 

Pauline and Tristan, would either of you be able to give it a whirl for your work?


t-downing (approved):
Yup, works with me! 

My only thought is whether to provide some backwards compatibility to process / load files that were created without the version number. But, for now it's probably easier to just rename the files (as opposed to re-download them, cause it takes so long). So happy with things as they are.

However, separately, there is an issue with loading the forecast data (which I don't think is covered in the tests). In comes up as well in the most recent release in so it's not related to this PR. I'll log it separately.

turnerm:
> My only thought is whether to provide some backwards compatibility to process / load files that were created without the version number. But, for now it's probably easier to just rename the files (as opposed to re-download them, cause it takes so long). So happy with things as they are.

Yes I considered that but I don't want to add more confusing input parameters, where it's only our team using it for now so we can easily just rename our files. 

> However, separately, there is an issue with loading the forecast data (which I don't think is covered in the tests). In comes up as well in the most recent release in so it's not related to this PR. I'll log it separately.

Great catch (and I know, I need to add load tests...), I threw a fix into this PR as well!

t-downing:
> Yes I considered that but I don't want to add more confusing input parameters, where it's only our team using it for now so we can easily just rename our files.

Yup makes sense!

> Great catch (and I know, I need to add load tests...), I threw a fix into this PR as well!

Amazing, thanks so much for doing that! I double checked and it works for me :)

turnerm:
Was hoping to hear back from Ervin before but he may be on holiday so I'll just merge for now.

turnerm:
Sorry @PaulineNimo @t-downing do you mind taking one last look

t-downing (approved):
Yup, working nicely for me! Thanks so much for getting this done.

t-downing:
> Oops, I wrote the above message before actually checking my email 🤦 The coordinate issues need to be fixed before merging, I'll try to do that asap. This means if you've used this code already for any v4 analysis then the rasters may have been interpolated and provide slightly reduced values. It probably won't affect your results qualitatively but once fixed you should probably do the download again, sorry about that.

Just for reference I didn't check if the coordinates issue actually changed things for me, but either way good to have the corrected version!

PaulineNimo (approved):
Great work!

turnerm:
> Not sure where to flag this but have been playing around with the data and it seems the new reanalysis version 4 data does not include the date "1979-01-01". The 1979 file only contains 364 time points. This is from the GRIB file downloaded from CDS. Everything else seems to download and process well with the file name now including the adjusted coordinates. I will check once I start loading the data if there are any challenges.

Thanks for noting that Pauline, I checked and indeed that is consistent with what's in CDS. A bit strange but I don't think it should break anything, particularly because the reanalysis is downloaded in year blocks.

